PR #15, opened 2023-11-24
Edit some text so that be more senseful.

HoytRen:
1. Because even the title of this cip is longer than 44 chars, I believe the unit of limit should be 'words'.

2. Adding a reference for relative rules could make it more senseful.

3. Do not need these rules that conflict with this cip itself.

Besides these modifications, after more thinking, I feel the Cip-1 should not be so complex. Just like the U.S. Constitution which has only rationale, rights, responsibility, and reference, but not the details of type, process, format, header, and editor. I believe these should be in separated cips because they could be changed more frequently.

### DIFF
--- a/cips/cip-1.md
+++ b/cips/cip-1.md
@@ -1,9 +1,9 @@
 ---
 cip: 1
 title: Celestia Improvement Proposal Process and Guidelines
+author: Yaz Khoury <yaz@celestia.org>
 status: Draft
 type: Meta
-author: Yaz Khoury <yaz@celestia.org>
 created: 2023-04-13
 ---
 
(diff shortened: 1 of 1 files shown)


cips/cip-1.md
```diff
@@ -311,9 +312,8 @@ format. There is a [CIP template](./cip-template.md) to follow.
 
 ## CIP Header Preamble
 
-Each CIP must begin with an RFC 822 style header preamble, preceded
-and followed by three hyphens (---). This header is also termed
-“front matter” by Jekyll. The headers must appear in the following order.
+Each CIP must begin with an RFC 822 style header preamble.
```
YazzyYaz, line 315: why remove the rule on hyphens? this is (---) how we construct the preamble. Observe the CIP template doc for example.

YazzyYaz (changes requested):

cips/cip-1.md
YazzyYaz: I won't approve this section.

HoytRen: Ok, I didn't convince you, then I reverted this section and changed the order of the preamble header instead.

YazzyYaz (approved):

PR #16, opened 2023-11-27
CIP: Standardize data expiry time for pruned nodes

musalbas:
[Rendered](https://github.com/musalbas/CIPs/blob/musalbas/cip-pruning/cips/cip-pruning-window.md)

Discussion: https://forum.celestia.org/t/cip-standardize-data-expiry-time-for-pruned-nodes/1326

### DIFF
--- a/cips/cip-pruning-window.md
+++ b/cips/cip-pruning-window.md
@@ -2,7 +2,7 @@
 title: Standardize data expiry time for pruned nodes
 description: Standardize default data expiry time for pruned nodes to 30 days.
 author: Mustafa Al-Bassam (@musalbas)
-discussions-to: URL
+discussions-to: https://forum.celestia.org/t/cip-standardize-data-expiry-time-for-pruned-nodes/1326
 status: Draft
 type: Standards Track
 category: Data Availability
--- a/cips/cip-pruning-window.md
+++ b/cips/cip-pruning-window.md
@@ -49,4 +49,3 @@ As discussed in Rationale, data availability sampling light nodes need to at lea
 ## Copyright
 
 Copyright and related rights waived via [CC0](../LICENSE).
-<!-- markdownlint-disable MD013 -->
--- a/cips/cip-pruning-window.md
+++ b/cips/cip-pruning-window.md
@@ -1,0 +1,52 @@
+---
+title: Standardize data expiry time for pruned nodes
+description: Standardize default data expiry time for pruned nodes to 30 days.
+author: Mustafa Al-Bassam (@musalbas)
+discussions-to: URL
+status: Draft
+type: Standards Track
+category: Data Availability
+created: 2023-11-23
+---
+
+## Abstract
+
+This CIP standardizes the default expiry time of historical blocks for pruned (non-archival) nodes to 30 days.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


cips/cip-pruning-window.md
```diff
@@ -0,0 +1,52 @@
...
+30 days is chosen for the following reasons:
+* Data availability sampling light nodes need to at least sample data within the Tendermint weak subjectivity period of 21 days in order to independently verify the data availability of the chain, and so they need to be able to sample data up to at least 21 days old.
+* 30 days ought to be a reasonable amount of time for data to be downloaded from the chain by any application that needs it.
+* EIP-4844 also proposes [30 days](https://notes.ethereum.org/@vbuterin/proto_danksharding_faq#If-data-is-deleted-after-30-days-how-would-users-access-older-blobs).
+
+## Backwards Compatibility
+
+On the Celestia data availability network, there is only one topic under which storage nodes can be discovered, where all nodes discovered under that topic are expected to be non-pruned nodes.
+
+An additional new topic will be created for pruned nodes, so that nodes discovered under the existing topic will not unexpectedly have missing block data when being queried.
```
musalbas, line 43: Should we name the exact topics here @Wondertan?

renaynay: You can name them `full` and `archival`

Where `full` is the default (pruned) and `archival` is historical

renaynay, line 43: The pruned node topic will not be an _additional_ topic but rather the current topic that we have which is `full`. The additional topic will be for archival nodes (`archival`).

We should call storage nodes `archival` to better indicate that they retain + serve historical blocks.

musalbas: As a side note, I was going to name them archival in this text, but then I saw that Ethereum refers to archival nodes as nodes that store all historical blocks but only prune state.

So I thought maybe it's better to be explicit and refer to them as block archival nodes; or non-pruned nodes. To avoid confusion.

But if you're saying `full` and `archival` are the actual topic names in libp2p, I don't think that matters.

musalbas: IMO it should be the other way round: the _existing_ topic should be for archival nodes, otherwise you will break backwards compatibility, because you will clutter the existing topic with non-archival nodes, which will not be backwards compatible with older nodes that are discovering nodes in the existing topic, and thinking they have all the historical blocks.

cips/cip-pruning-window.md
```diff
@@ -0,0 +1,52 @@
+---
+title: Standardize data expiry time for pruned nodes
+description: Standardize default data expiry time for pruned nodes to 30 days.
+author: Mustafa Al-Bassam (@musalbas)
+discussions-to: URL
```
YazzyYaz, line 5: Needs a discussion link. Maybe open one in the forum

cips/cip-pruning-window.md
```diff
@@ -0,0 +1,52 @@
...
+An additional new topic will be created for pruned nodes, so that nodes discovered under the existing topic will not unexpectedly have missing block data when being queried.
+
+## Security Considerations
+
+As discussed in Rationale, data availability sampling light nodes need to at least sample data within the Tendermint weak subjectivity period of 21 days in order to independently verify the data availability of the chain. 30 days exceeds this.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE).
+<!-- markdownlint-disable MD013 -->
```
YazzyYaz, line 52: can delete this line

musalbas: What's the implications of deleting that line? Should that also be removed from the template then ?

musalbas:
Moving discussion to https://forum.celestia.org/t/cip-standardize-data-expiry-time-for-pruned-nodes/1326

YazzyYaz (approved):

PR #29, opened 2023-12-07
cip for coordinated network upgrades

cmwaters:
Adds a draft CIP defining a protocol for coordinated upgrades

### DIFF
--- a/cips/cip-network-upgrade.md
+++ b/cips/cip-network-upgrade.md
@@ -34,7 +34,7 @@ message MsgSignalVersion {
 message MsgTryUpgrade { string signer = 1; }
 ```
 
-Only validators can submit `MsgSignalVersion` . The Celestia state machine tracks which version each validator has signalled for. The signalled version MUST either be the same version or the next. There is no support for skipping versions or downgrading.
+Only validators can submit `MsgSignalVersion`. The Celestia state machine tracks which version each validator has signalled for. The signalled version MUST either be the same version or the next. There is no support for skipping versions or downgrading.
 
 Clients may query the tally for each version as follows:
 
--- a/cips/cip-network-upgrade.md
+++ b/cips/cip-network-upgrade.md
@@ -66,6 +66,10 @@ This feature modifies the functionality of the state machine in a breaking way a
 
 As the API is additive, there is no need to consider backwards compatibility for clients.
 
+## Reference Implementation
+
+TBA
+
 ## Security Considerations
 
 See the section on rationale for understanding the network halting risk.
--- a/cips/cip-network-upgrade.md
+++ b/cips/cip-network-upgrade.md
@@ -1,0 +1,75 @@
+---
+title: Coordinated network upgrades
+description: Protocol for coordinating major network upgrades
+author: Callum Waters (@cmwaters)
+discussions-to: URL
+status: Draft
+type: Standards Track
+category: Core
+created: 2023-12-7
+---
+
+## Abstract
+
+Use a pre-programmed height for the next major upgrade. Following major upgrades will use an in-protocol signalling mechanism. Validators will submit messages to signal their ability and preference to use the next version. Once a quorum of 5/6 has signalled the same version, the network will migrate to that version.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


cips/cip-network-upgrade.md
```diff
@@ -0,0 +1,75 @@
...
+
+If the network agrees to move to a version that is not supported by the node, the node will gracefully shutdown.
+
+## Rationale
+
+When attempting a major upgrade, there is increased risk that the network halts. At least 2/3 in voting power of the network is needed to migrate and agree upon the next block. However this does not take into account the actions of byzantine validators. As an example, say there is 100 in total voting power. 66 have signalled v2 and 33 have yet to signal i.e. they are still signalling v1. It takes 1 byzantine voting power to pass the threshold, convincing the network to propose v2 blocks and then omitting their vote leaving the network failing to reach consensus until one of the 33 are able to upgrade. At the other end of the spectrum, increasing the necessary threshold means less voting power required to veto the upgrade. The middle point here is setting a quorum of 5/6ths which provides 1/6 byzantine fault tolerance to liveness and requiring at least 1/6th of the network to veto the upgrade.
```
evan-forbes, line 57: this is really good rationale for why the percentage is over 2/3s, but doesn't contian the rest of the rational for upgrades that don't use token voting or single binaries. I'm not blocking on it, but we might want to include more here from previous documents

evan-forbes (approved):
:fire:

cips/cip-network-upgrade.md
```diff
@@ -0,0 +1,75 @@
+---
+title: Coordinated network upgrades
+description: Protocol for coordinating major network upgrades
+author: Callum Waters (@cmwaters)
+discussions-to: URL
```
YazzyYaz, line 5: post a URL please from forum for discussions

cips/cip-network-upgrade.md
```diff
@@ -0,0 +1,75 @@
...
+
+An on-chain tallying system was decided over an off-chain as it canonicalises the information which is important for coordination and adds accountability to validators. As a result of this decision, validators will need to pay fees to upgrade which is necessary to avoid spamming the chain.
+
+## Backwards Compatibility
+
+This feature modifies the functionality of the state machine in a breaking way as the state machine can now dictate version changes. This will require a major upgrade to implement (thus the protocol won’t come into affect until the following major upgrade).
+
+As the API is additive, there is no need to consider backwards compatibility for clients.
+
+## Security Considerations
```
YazzyYaz, line 69: Is there a Reference Implementation section?

cips/cip-network-upgrade.md
YazzyYaz: I would recommend discussions on the proposal to happen in a forum post that @cmwaters should provide. PRs get merged so convos get lost over time.

evan-forbes: SGTM and makes sense

cips/cip-network-upgrade.md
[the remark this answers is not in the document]
cmwaters: They use their operator account key (not their cometbft consensus key). It is not automatic, however you could have something setup with authz that permitted the node to automatically sign the signal version transaction. I think this is outside the scope of the CIP which should be focused on the specification change to the state machine

cips/cip-network-upgrade.md
cmwaters: Done although it's a bit of a chicken and egg problem. I will update the forum discussion to link to the actual document instead of the PR for the document after merging this

cips/cip-network-upgrade.md
YazzyYaz: yes you can. I just meant having the section here, even if the implementation isn't ready. can just add to it "TBA" or something. so far, there's no Implementation section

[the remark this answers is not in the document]
rootulp: Agreed out of scope. Resolving

cmwaters: Done

YazzyYaz (approved):

PR #41, opened 2024-01-08
docs: move CIP-11 to withdrawn

rootulp:
We're putting the refund unspent gas line of work on pause because the mitigation strategies to the DoS attack are overly complicated.


cips/cip-11.md
```diff
@@ -52,17 +52,18 @@ TBA
 
 ## Reference Implementation
 
-TBA
+<https://github.com/celestiaorg/celestia-app/pull/2887>
```
rootulp, line 55: I think there's an effort to minimize the number of links to external sources. If we'd rather not include this link here, please LMK. I can copy + paste a portion of the implementation here but that seemed less preferable than a link to the full PR which contains more context.

jcstein (approved):
LGTM

PR #52, opened 2024-01-24
feat: add auto-reviewers

jcstein:
## Overview

Adds auto-review workflow to tag editors for review on PRs.

## Checklist


.github/workflows/housekeeping.yml
```diff
@@ -0,0 +1,19 @@
+name: Housekeeping
```
rootulp, line 1: [nit] I think this has been copied from some other celestiaorg repos but I would prefer renaming to be more descriptive.

```suggestion
name: pr-review-requester
```

In addition, we may consider renaming `housekeeping.yml` to `pr-review-requester.yml`.

## Motivation
The "housekeeping" workflow name communicates very little about what this workflow does. This is especially problematic because the Github actions sidebar retains old workflow names for a while so it can be confusing what some of these do when they don't have descriptive names. 

Screenshot 2024-01-24 at 5 00 56 PM

Ref: https://github.com/celestiaorg/celestia-app/blob/b1592c314d4dca77c477e62ee1890970c6ee6e0c/.github/workflows/pr-review-requester.yml#L1

jcstein: SGTM, thanks for the review @rootulp

rootulp (approved):

.github/workflows/pr-review-requester.yml
```diff
@@ -0,0 +1,19 @@
...
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened, ready_for_review]
+
+jobs:
+    auto-add-reviewer:
+        name: Auto add reviewer to PR
+        if: github.event.pull_request
+        uses: celestiaorg/.github/.github/workflows/reusable_housekeeping.yml@v0.2.2
```
MSevey, line 13:
```suggestion
        uses: celestiaorg/.github/.github/workflows/reusable_housekeeping.yml@v0.3.1
```

.github/workflows/pr-review-requester.yml
```diff
@@ -0,0 +1,19 @@
+name: pr-review-requester
+
+on:
+  issues:
+    types: [opened]
```
MSevey, line 5: The auto request review doesn't do anything for issues. 
There are some other actions that can help with that if you want. 
But for this you can just remove this.

.github/workflows/pr-review-requester.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request_target:
+    types: [opened, ready_for_review]
+
+jobs:
+    auto-add-reviewer:
+        name: Auto add reviewer to PR
+        if: github.event.pull_request
```
MSevey, line 12: if you remove the issue trigger, you can remove this.

.github/workflows/pr-review-requester.yml
```diff
@@ -0,0 +1,19 @@
...
+        uses: celestiaorg/.github/.github/workflows/reusable_housekeeping.yml@v0.2.2
+        secrets: inherit
+        permissions:
+          issues: write
```
MSevey, line 16:
```suggestion
```
don't need issue permissions for this action.

MSevey (changes requested):
small updates.

.github/workflows/pr-review-requester.yml
jcstein: dacf27e36da6dfadd754588e204a4fc5152838e7 done! thank you

.github/workflows/pr-review-requester.yml
jcstein: e709df605e8bb6b6bc97a12a1bbbdfa5f8e6ff0a and done, thanks again for the review

YazzyYaz (approved):

MSevey (approved):

jcstein:
@MSevey can you please help debug [this error on the action](https://github.com/celestiaorg/CIPs/actions/runs/7704744309)?

> The workflow is not valid. .github/workflows/pr-review-requester.yml (Line: 8, Col: 5): Error calling workflow 'celestiaorg/.github/.github/workflows/reusable_housekeeping.yml@v0.3.1'. The nested job 'labels' is requesting 'issues: write', but is only allowed 'issues: none'.

jcstein:
We will also need to modify github settings to allow `ebuchman` as a reviewer, based on #56

MSevey:
> @MSevey can you please help debug [this error on the action](https://github.com/celestiaorg/CIPs/actions/runs/7704744309)?
> 
> > The workflow is not valid. .github/workflows/pr-review-requester.yml (Line: 8, Col: 5): Error calling workflow 'celestiaorg/.github/.github/workflows/reusable_housekeeping.yml@v0.3.1'. The nested job 'labels' is requesting 'issues: write', but is only allowed 'issues: none'.

@jcstein 

```
The workflow is not valid. .github/workflows/pr-review-requester.yml (Line: 8, Col: 5): Error calling workflow 'celestiaorg/.github/.github/workflows/reusable_housekeeping.yml@v0.3.1'. The nested job 'labels' is requesting 'issues: write', but is only allowed 'issues: none'.
```
Looks like I was mistaken. Even though the review action doesn't require issue permissions, because there is another step in the workflow that does, it needs to be there. 

so you just need to add `issues: write` back under permissions.

jcstein:
Using this thread as it's the most contained set of info. Now, I think it may be permissions-related?

https://github.com/celestiaorg/CIPs/actions/runs/7713106957

> The workflow is not valid. .github/workflows/pr-review-requester.yml (Line: 8, Col: 5): Error calling workflow 'celestiaorg/.github/.github/workflows/reusable_housekeeping.yml@v0.3.1'. The nested job 'labels' is requesting 'issues: write', but is only allowed 'issues: none'.

PR #63, opened 2024-02-01
bug: rename auto-request-review.yml to auto_request_review.yml

MSevey:
## Overview

The use of `-` vs `_` in the name was missed on the original PR. Updating to fix the error of not seeing the config file.


rootulp (approved):
Looks like the default is `auto_request_review.yml`

Ref: https://github.com/necojackarc/auto-request-review/blob/d79257da48ade29bb20ff86bbd90fd3432e88255/README.md?plain=1#L184

MSevey:
> Looks like the default is `auto_request_review.yml`
> 
> Ref: https://github.com/necojackarc/auto-request-review/blob/d79257da48ade29bb20ff86bbd90fd3432e88255/README.md?plain=1#L184

correct, thanks for linking the docs 👍

jcstein (approved):
Thanks @MSevey and @rootulp! LGTM

PR #68, opened 2024-02-06
fix: gov changeable param typo

caomingpei:
### Commit Summary:

Fix Typo in CIP-13 Regarding *consensus.evidence.MaxAgeDuration*

### Commit Message:

This commit addresses a documentation error in cips/cip-13.md concerning the *consensus.evidence.MaxAgeDuration* parameter.

The current documentation incorrectly suggests that *consensus.evidence.MaxAgeDuration* is changeable via governance. This is a discrepancy because *consensus.evidence.MaxAgeDuration* should be identical with the *staking.UnbondingTime*, which is not changeable on-chain as indicated by its governance changeability status being set to False.

**Additional comment:** Allowing *consensus.evidence.MaxAgeDuration* to be modified on-chain poses a security risk. Such flexibility could enable a node to reduce this duration to under 3 days, which is problematic. In such a scenario, evidence older than this reduced period (less than 3 days) would be deemed invalid, potentially opening a attack window for fraud proofs to be submitted to the rollup. (Especially, the common setting of challenge period of rollup is 3-14 days).

To maintain consistency with *staking.UnbondingTime*, the parmeter *consensus.evidence.MaxAgeDuration* should remain unchanged by on-chain governance. This commit fixes the typo.

## Checklist


rootulp:
We shouldn't merge this PR because it isn't a typo. I agree the param likely shouldn't be changeable by governance but it currently is changeable by governance. See [this test](https://github.com/celestiaorg/celestia-app/blob/5f39a36358bbe89cb564448deac3cf5a95e5786f/x/paramfilter/test/gov_params_test.go#L191-L193).

Instead of this PR, we should make a CIP to convert this param from governance modifiable to unmodifiable.

jcstein:
Agreed on closing this PR.[ The celestia-app specifications also list it as one that is modifiable by governance](https://github.com/celestiaorg/celestia-app/blob/4daf069ad1ff030c040ded077ca4922af057d91c/specs/src/specs/params.md?plain=1#L33).

caomingpei:
> Contributor
> ###

Yeah you are right, I agree with you. Simply modifying this CIP would cause the implementation not to match the protocol. 
I am interested in these settings (such as *consensus.evidence.MaxAgeDuration* and *consensus.evidence.MaxAgeNumBlocks*), because I realized that these settings may introduce security risk (Long Range Attack for PoS). Is there any chance we can come up with a new CIP on this issue?

jcstein:
Yes, absolutely. Please feel free to do so in the forum: https://forum.celestia.org/c/celestia-improvement-proposal-cip/31

jcstein:
what is your ID on the forum @caomingpei ?

PR #74, opened 2024-02-07
docs: CIP for make security related governance parameters immutable

caomingpei:
Add a CIP to make two parameters unchangeable for online governance.

Motivation: [celestiaorg/CIPs/pull#68](https://github.com/celestiaorg/CIPs/pull/68)

Forum: https://forum.celestia.org/t/cip-make-security-related-governance-parameters-immutable/1566

### DIFF
--- a/cips/cip-draft_msrgpi_abbrev.md
+++ b/cips/cip-draft_msrgpi_abbrev.md
@@ -1,4 +1,5 @@
 ---
+cip: 16
 title: Make Security Related Governance Parameters Immutable
 description: Consensus-related parameters should not be modified via on-chain governance in the Celestia network.
 author: Mingpei CAO (@caomingpei)
--- a/cips/cip-draft_msrgpi_abbrev.md
+++ b/cips/cip-draft_msrgpi_abbrev.md
@@ -20,7 +20,6 @@ Consensus protocols play an important role in Celestia. As a Data Availability s
 
 ## Specification
 
-
 |  Param \ Changeable via Governance   | Current | **Proposed**  |
 |:------------------------------------:|:-------:|:-------------:|
 | `consensus.evidence.MaxAgeDuration`  |  True   |   **False**   |
@@ -33,7 +32,6 @@ Consensus protocols play an important role in Celestia. As a Data Availability s
 Adopting an on-chain governance method comes with inherent risks of governance attacks, particularly concerning parameters related to `consensus.evidence`.
 As outlined in the [decentralized governance documentation](https://docs.celestia.org/learn/staking-governance-supply), all holders of the native token TIA can propose and vote on on-chain governance proposals. It is unrealistic to expect the majority of holders to understand the technical details of the protocol and implementation thoroughly. Consequently, on-chain governance participants (a.k.a. TIA holders) may be incentivized (or, in some cases, bribed) to vote for or against a proposal without understanding the potential impact. Any changeable parameter within the `Governance Parameter for Celestia` could be targeted for changes through on-chain governance proposals. Therefore, making security-related parameters unchangeable for the on-chain governance proposal could serve as an effective solution to alleviate introduced risks.
 
-
 ### Inconsistency
 
 | Module.Parameter                     | Default                    | Summary                                                                                                                             | Changeable via Governance |
@@ -44,8 +42,6 @@ As outlined in the [decentralized governance documentation](https://docs.celesti
 | `staking.UnbondingTime`              | 1814400 (21 days)          | Duration of time for unbonding in seconds.                                                                                          | False                     |
 
 This is a part of the table introduced in [CIP-13](cip-13.md). The summary of parameter `consensus.evidence.MaxAgeDuration` states *"...This value should be identical to the unbonding period"*. Meanwhile, the parameter `staking.UnbondingTime` is NOT changeable since the `Changeable via Governance` is set to False. Suppose an on-chain governance proposal tries to modify the default value of `consensus.evidence.MaxAgeDuration` from `1814400000000000 (21 days)` to a different value. It would create an inconsistency between the description and implementation because the modified value would no longer be identical to the unbonding period.
-
-
 
 ### Security Risk
 
@@ -64,16 +60,13 @@ Those two parameters are used in the function of `verify.go`, which is responsib
 
 Additionally, suppose that an on-chain governance proposal sets the `evidence.MaxAgeDuration` and `evidence.MaxAgeNumBlocks` to extremely low values, meaning that evidence expires quickly. If a malicious validator were to engage in Duplicate Vote or Light Client Attack, it would lead to consensus instability. Given that Celestia is a solution data availability, this consensus instability would introduce security risk to the upper-layer applications (e.g. rollup). A detailed discussion can be found in the Security Considerations.
 
-
 ### Summary
 
 In summary, configuring these two parameters as immutable values that can NOT be changed via on-chain governance can mitigate the risks of inconsistency and security issues introduced by unintentional (or malicious) governance activities. Moreover, in the face of a security incident that concerns these parameters, reliance on on-chain governance may be inadequate. Implementing modifications through a hard fork represents a more resilient approach.
 
-
 ## Backwards Compatibility
 
 It maintains backward compatibility provided that NO on-chain governance proposal to change the two parameters has been approved at the time of this CIP implementation.
-
 
 ## Reference Implementation
 
@@ -92,13 +85,11 @@ The above example serves as a conceptual illustration, and `MaxBytes` in `ParamS
 
 Besides, relevant documents should be updated accordingly, such as [Celestia App Specifications](https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/params.md)
 
-
 ## Security Considerations
 
 This CIP recommends setting those two parameters as immutable constants that are NOT allowed to change via on-chain governance proposals. Adopting this CIP means future changes about those two parameters require community coordination and hard forks. Although hard forks carry the community-divided risk, it is worth noting that many blockchain communities have successfully navigated multiple hard forks. The risks of division and disagreement can be minimized by having thorough discussions and working towards widespread agreement before moving forward with a hard fork. Consequently, the risk is manageable and should not be a significant concern.
 
 However, if modifications to those two parameters via on-chain governance are allowed to persist, this could not only result in the inconsistency between the protocol and implementation but also introduce the following potential security risks:
-
 
 ### Consensus Instability
 
@@ -124,11 +115,9 @@ However, if modifications to those two parameters via on-chain governance are al
 
 Assuming that `evidence.MaxAgeDuration` and `evidence.MaxAgeNumBlocks` are configured with extremely low values, such a configuration implies that evidence would expire rapidly. Under these conditions, malicious validators might exploit the system by double voting on two blocks simultaneously, potentially leading to temporary network partitions. In the context of a Data Availability solution like Celestia, any instability in consensus could compromise the blob transactions. This situation risks the security and reliability of upper-layer rollups dependent on this infrastructure.
 
-
 ### Resource Consumption
 
 Resource consumption attacks targeting ComeBFT consensus algorithms aim to exploit the resource constraints of nodes within the network. The summary of `consensus.evidence.MaxAgeNumBlocks` states *"...This value will stop CometBFT from pruning block data"*. If the pruned window is set too large, nodes must store all the relevant data for the extended period, which can inundate their storage and memory capacity.
-
 
 ## Copyright
 
--- a/cips/cip-draft_msrgpi_abbrev.md
+++ b/cips/cip-draft_msrgpi_abbrev.md
@@ -66,7 +66,7 @@ In summary, configuring these two parameters as immutable values that can NOT be
 
 ## Backwards Compatibility
 
-It maintains backward compatibility provided that NO on-chain governance proposal to change the two parameters has been approved at the time of this CIP implementation.
+This CIP recommends freezing on-chain governance for two security related parameters, which introduces backward incompatibility. This incompatibility is due to any future modifications to these parameters requiring at least a hardfork.
 
 ## Reference Implementation
 
@@ -87,9 +87,9 @@ Besides, relevant documents should be updated accordingly, such as [Celestia App
 
 ## Security Considerations
 
-This CIP recommends setting those two parameters as immutable constants that are NOT allowed to change via on-chain governance proposals. Adopting this CIP means future changes about those two parameters require community coordination and hard forks. Although hard forks carry the community-divided risk, it is worth noting that many blockchain communities have successfully navigated multiple hard forks. The risks of division and disagreement can be minimized by having thorough discussions and working towards widespread agreement before moving forward with a hard fork. Consequently, the risk is manageable and should not be a significant concern.
+This CIP recommends setting those two parameters as immutable constants which are NOT allowed to change via on-chain governance proposals. Adopting this CIP means future changes for those two parameters require community coordination and hardforks. Although hardforks carry the community-divided risk, it is worth noting that many blockchain communities have successfully navigated multiple hardforks. The risks of division and disagreement can be minimized by thoroughly discussing and working towards widespread agreement before moving forward with a hardfork. Consequently, the risk is manageable and should not be a significant concern.
 
-However, if modifications to those two parameters via on-chain governance are allowed to persist, this could not only result in the inconsistency between the protocol and implementation but also introduce the following potential security risks:
+If the proposed two parameters have not been changed since genesis, the security impact of making them NOT changeable via on-chain governance is not significant. Besides, if modifications to those two parameters via on-chain governance are still allowed, this could not only result in the inconsistency between the protocol and implementation but also introduce the following potential security risks:
 
 ### Consensus Instability
 
--- a/cips/cip-draft_msrgpi_abbrev.md
+++ b/cips/cip-draft_msrgpi_abbrev.md
@@ -58,7 +58,7 @@ func (evpool *Pool) verify(evidence types.Evidence) error {
 
 Those two parameters are used in the function of `verify.go`, which is responsible for verifying whether evidence has expired or not. According to the `if` statement, even without modifying `consensus.evidence.MaxAgeDuration`, it is still possible to prolong the expiration time of evidence by increasing `consensus.evidence.MaxAgeNumBlocks`, which means that older evidence will be considered valid. An extended expiration time of evidence introduces the potential risk of Resource Consumption, and a detailed discussion can be found in the Security Considerations. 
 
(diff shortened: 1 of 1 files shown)


cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+## Abstract
+
+This CIP suggests that the security-critical governance parameters—`consensus.evidence.MaxAgeDuration` and `consensus.evidence.MaxAgeNumBlocks`—should be immutable to on-chain governance proposals. In light of the present Celestia specifications and the details provided in [CIP-13](cip-13.md), maintaining the mutability of those parameters could open the door to future on-chain proposals that alter their values, potentially leading to an inconsistency between the protocol and implementation. This CIP also briefly analyzes the potential security risks that may arise from the parameter modifications.
+
+## Motivation
+
+Consensus protocols play an important role in Celestia. As a Data Availability solution, the robustness of Celestia is crucial to the performance and reliability of higher-level applications, such as rollups. [CIP-13](cip-13.md) introduces a framework for the on-chain governance parameters. Considering that governance proposals have the potential to alter parameter values, it is essential to designate certain parameters as immutable for the on-chain governance to prevent inconsistencies and mitigate security risks.
+
+## Specification
+
```
jcstein, line 22:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+## Rationale
+
+Adopting an on-chain governance method comes with inherent risks of governance attacks, particularly concerning parameters related to `consensus.evidence`.
+As outlined in the [decentralized governance documentation](https://docs.celestia.org/learn/staking-governance-supply), all holders of the native token TIA can propose and vote on on-chain governance proposals. It is unrealistic to expect the majority of holders to understand the technical details of the protocol and implementation thoroughly. Consequently, on-chain governance participants (a.k.a. TIA holders) may be incentivized (or, in some cases, bribed) to vote for or against a proposal without understanding the potential impact. Any changeable parameter within the `Governance Parameter for Celestia` could be targeted for changes through on-chain governance proposals. Therefore, making security-related parameters unchangeable for the on-chain governance proposal could serve as an effective solution to alleviate introduced risks.
+
+
```
jcstein, line 36:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+| ......                               | ......                     | ......                                                                                                                              | ......                    |
+| `staking.UnbondingTime`              | 1814400 (21 days)          | Duration of time for unbonding in seconds.                                                                                          | False                     |
+
+This is a part of the table introduced in [CIP-13](cip-13.md). The summary of parameter `consensus.evidence.MaxAgeDuration` states *"...This value should be identical to the unbonding period"*. Meanwhile, the parameter `staking.UnbondingTime` is NOT changeable since the `Changeable via Governance` is set to False. Suppose an on-chain governance proposal tries to modify the default value of `consensus.evidence.MaxAgeDuration` from `1814400000000000 (21 days)` to a different value. It would create an inconsistency between the description and implementation because the modified value would no longer be identical to the unbonding period.
+
+
```
jcstein, line 48:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+| ......                               | ......                     | ......                                                                                                                              | ......                    |
+| `staking.UnbondingTime`              | 1814400 (21 days)          | Duration of time for unbonding in seconds.                                                                                          | False                     |
+
+This is a part of the table introduced in [CIP-13](cip-13.md). The summary of parameter `consensus.evidence.MaxAgeDuration` states *"...This value should be identical to the unbonding period"*. Meanwhile, the parameter `staking.UnbondingTime` is NOT changeable since the `Changeable via Governance` is set to False. Suppose an on-chain governance proposal tries to modify the default value of `consensus.evidence.MaxAgeDuration` from `1814400000000000 (21 days)` to a different value. It would create an inconsistency between the description and implementation because the modified value would no longer be identical to the unbonding period.
+
```
jcstein, line 47:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+        ...
+    }
+    ...
+}
+```
+
+Those two parameters are used in the function of `verify.go`, which is responsible for verifying whether evidence has expired or not. According to the `if` statement, even without modifying `consensus.evidence.MaxAgeDuration`, it is still possible to prolong the expiration time of evidence by increasing `consensus.evidence.MaxAgeNumBlocks`, which means that older evidence will be considered valid. An extended expiration time of evidence introduces the potential risk of Resource Consumption, and a detailed discussion can be found in the Security Considerations. 
+
+Additionally, suppose that an on-chain governance proposal sets the `evidence.MaxAgeDuration` and `evidence.MaxAgeNumBlocks` to extremely low values, meaning that evidence expires quickly. If a malicious validator were to engage in Duplicate Vote or Light Client Attack, it would lead to consensus instability. Given that Celestia is a solution data availability, this consensus instability would introduce security risk to the upper-layer applications (e.g. rollup). A detailed discussion can be found in the Security Considerations.
+
```
jcstein, line 66:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Summary
+
+In summary, configuring these two parameters as immutable values that can NOT be changed via on-chain governance can mitigate the risks of inconsistency and security issues introduced by unintentional (or malicious) governance activities. Moreover, in the face of a security incident that concerns these parameters, reliance on on-chain governance may be inadequate. Implementing modifications through a hard fork represents a more resilient approach.
+
```
jcstein, line 71:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Backwards Compatibility
+
+It maintains backward compatibility provided that NO on-chain governance proposal to change the two parameters has been approved at the time of this CIP implementation.
+
```
jcstein, line 76:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
++       // consensus.evidence.MaxAgeDuration and .MaxAgeNumBlocks
++       {baseapp.Paramspace, string(baseapp.ParamStoreKeyEvidenceParams)},		
+	}
+}
+```
+The above example serves as a conceptual illustration, and `MaxBytes` in `ParamStoreKeyEvidenceParams` should still remain changeable.
+
+Besides, relevant documents should be updated accordingly, such as [Celestia App Specifications](https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/params.md)
+
+
```
jcstein, line 95:
```suggestion
```

jcstein, line 95:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+## Security Considerations
+
+This CIP recommends setting those two parameters as immutable constants that are NOT allowed to change via on-chain governance proposals. Adopting this CIP means future changes about those two parameters require community coordination and hard forks. Although hard forks carry the community-divided risk, it is worth noting that many blockchain communities have successfully navigated multiple hard forks. The risks of division and disagreement can be minimized by having thorough discussions and working towards widespread agreement before moving forward with a hard fork. Consequently, the risk is manageable and should not be a significant concern.
+
+However, if modifications to those two parameters via on-chain governance are allowed to persist, this could not only result in the inconsistency between the protocol and implementation but also introduce the following potential security risks:
+
+
```
jcstein, line 102:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+                            |
+                            |   +---------------------+
+                            ·—— | Block Header (H1')  |
+                                +---------------------+   ......
+                                | PayForBlobs'        |
+                                +---------------------+
+```
+
+Assuming that `evidence.MaxAgeDuration` and `evidence.MaxAgeNumBlocks` are configured with extremely low values, such a configuration implies that evidence would expire rapidly. Under these conditions, malicious validators might exploit the system by double voting on two blocks simultaneously, potentially leading to temporary network partitions. In the context of a Data Availability solution like Celestia, any instability in consensus could compromise the blob transactions. This situation risks the security and reliability of upper-layer rollups dependent on this infrastructure.
+
```
jcstein, line 126:
```suggestion
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Resource Consumption
+
+Resource consumption attacks targeting ComeBFT consensus algorithms aim to exploit the resource constraints of nodes within the network. The summary of `consensus.evidence.MaxAgeNumBlocks` states *"...This value will stop CometBFT from pruning block data"*. If the pruned window is set too large, nodes must store all the relevant data for the extended period, which can inundate their storage and memory capacity.
+
+
```
jcstein, line 132:
```suggestion
```

jcstein (changes requested):

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,124 @@
...
+| ......                               | ......                     | ......                                                                                                                              | ......                    |
+| `staking.UnbondingTime`              | 1814400 (21 days)          | Duration of time for unbonding in seconds.                                                                                          | False                     |
+
+This is a part of the table introduced in [CIP-13](cip-13.md). The summary of parameter `consensus.evidence.MaxAgeDuration` states *"...This value should be identical to the unbonding period"*. Meanwhile, the parameter `staking.UnbondingTime` is NOT changeable since the `Changeable via Governance` is set to False. Suppose an on-chain governance proposal tries to modify the default value of `consensus.evidence.MaxAgeDuration` from `1814400000000000 (21 days)` to a different value. It would create an inconsistency between the description and implementation because the modified value would no longer be identical to the unbonding period.
```
rootulp, line 44: [clarifying question] I think our specs are actually wrong here. I think the MaxAgeDuration doesn't need to be *identical* to the unbonding time but just >= unbonding time. For example, could the unbonding time be set to 30 days?

Update: I see the point about resource consumption below. Given evidence is usually empty, the resource consumption on nodes for a longer unboding time seems negligible to me.

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,124 @@
...
+        ...
+    }
+    ...
+}
+```
+
+Those two parameters are used in the function of `verify.go`, which is responsible for verifying whether evidence has expired or not. According to the `if` statement, even without modifying `consensus.evidence.MaxAgeDuration`, it is still possible to prolong the expiration time of evidence by increasing `consensus.evidence.MaxAgeNumBlocks`, which means that older evidence will be considered valid. An extended expiration time of evidence introduces the potential risk of Resource Consumption, and a detailed discussion can be found in the Security Considerations. 
+
+Additionally, suppose that an on-chain governance proposal sets the `evidence.MaxAgeDuration` and `evidence.MaxAgeNumBlocks` to extremely low values, meaning that evidence expires quickly. If a malicious validator were to engage in Duplicate Vote or Light Client Attack, it would lead to consensus instability. Given that Celestia is a solution data availability, this consensus instability would introduce security risk to the upper-layer applications (e.g. rollup). A detailed discussion can be found in the Security Considerations.
```
rootulp, line 61: [nit] the grammar here seems incorrect. 

```suggestion
Additionally, suppose that an on-chain governance proposal sets the `evidence.MaxAgeDuration` and `evidence.MaxAgeNumBlocks` to extremely low values, meaning that evidence expires quickly. If a malicious validator were to engage in Duplicate Vote or Light Client Attack, it would lead to consensus instability. Given that Celestia is a solution to data availability, this consensus instability would introduce security risk to the upper-layer applications (e.g. rollup). A detailed discussion can be found in the Security Considerations.
```

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Summary
+
+In summary, configuring these two parameters as immutable values that can NOT be changed via on-chain governance can mitigate the risks of inconsistency and security issues introduced by unintentional (or malicious) governance activities. Moreover, in the face of a security incident that concerns these parameters, reliance on on-chain governance may be inadequate. Implementing modifications through a hard fork represents a more resilient approach.
+
+## Backwards Compatibility
+
+It maintains backward compatibility provided that NO on-chain governance proposal to change the two parameters has been approved at the time of this CIP implementation.
```
rootulp, line 69: [question] I think this CIP is actually backwards incompatible because a change needs to be made to celestia-app to make these params unmodifiable. So this proposal needs to be introduced via a hard-fork.

cips/cip-draft_msrgpi_abbrev.md
caomingpei: This is a good question. It is well known that setting `evidence.maxAge` shorter than the `unbonding period` could introduce security risks. In the previous description, setting both values as equal seems reasonable. Although a longer time will not consume too much storage space, the meaning of evidence is lost if it is longer than the unbonding period. Suppose there is an attack, even if the evidence is still valid after a period longer than the unbonding period, but the attacker has already unstaked. Different value settings in evidence expiration and unbonding periods could also lead to confusion. It is unclear which value should be chosen for the "challenge period" for users. If the evidence always equals the unbonding period, we can also simplify the Celestia description, i.e., we need to confirm the validity of consensus within the unbonding period.

cips/cip-draft_msrgpi_abbrev.md
caomingpei: Thanks for your help!

cips/cip-draft_msrgpi_abbrev.md
caomingpei: The section I wrote needs to be clearer. What I meant to convey is that if the two parameters have not been modified, the setting for the validity period of `evidence` is still compatible after being banned from on-chain governance. However, I did not clearly state that the client proposing on-chain governance is no longer backwards compatible. My current thought is to amend this content to be backwards incompatible. And add a note explaining that if the parameters have never been changed, the security impact on consensus is not significant. Do you have any suggestions or comments more?

rootulp: > My current thought is to amend this content to be backwards incompatible.

Agreed.

> And add a note explaining that if the parameters have never been changed, the security impact on consensus is not significant.

Hmm I think this may belong in the [Security Considerations](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-template.md#security-considerations) section rather than here in **Backwards Compatibility**. Thanks for your thorough CIP and attention to detail!

cips/cip-draft_msrgpi_abbrev.md
rootulp: Thanks for your thorough response. It seems like we want evidnece.MaxAgeDuration to == unbonding period for simplicity and because the protocol doesn't _need_ to retain evidence longer than the unbonding period because a malicious validator could unbond and exit the protocol in that time. Seems reasonable to me.

If we wanted to provide a bit more info in this section, we could analyze the MaxAgeDuration and unboding period of other Cosmos chains (like Cosmos Hub, Osmosis). Something like:


Chain      | MaxAgeDuration             | Unbonding time
-----------|----------------------------|------------------
Celestia   | 1814400000000000 (21 days) | 1814400 (21 days)
Cosmos Hub | TBD                           | 21 days
Osmosis    |  TBD                          | 14 days

and then if they're identical (on a per-chain basis), that could be supporting evidence for making the param governance unmodifiable and exactly match the unbonding time.

caomingpei: |  **Chain**  | **MaxAgeDuration** |**Unbonding time**|
|:------------------------------------:|:-------:|:-------------:|
| Celestia | 1814400000000000 (21 days) |  1814400 (21 days) |
| Cosmos Hub |  *Complex Situation*\*   |   21 days  |
| Osmosis | 14 days (UnbondingTime) | 14 days |

The [MaxAgeDuration](https://github.com/osmosis-labs/osmosis/blob/be6d52021cde0e07e55ed071945ccb791bfe23dd/cmd/osmosisd/cmd/genesis.go#L495) of Osmosis is the same as its unbonding time.

However, for Cosmos Hub,  the situation is complex. For the genesis, the [max_evidence_age](https://github.com/cosmos/gaia/blob/c5de5d08816474c15a55a91c4a06d755805440a9/docs/docs/resources/genesis.md?plain=1#L301) in slashing is set to 1814400000000000 (21 days), which equals to the unbonding time. Moreover, this [parameter](https://github.com/cosmos/gaia/blob/c5de5d08816474c15a55a91c4a06d755805440a9/docs/docs/governance/proposal-types/params-change/Slashing.mdx#L151) remains unchanged until the genesis of `cosmoshub-4`. We can check the [#PR5952](https://github.com/cosmos/cosmos-sdk/pull/5952/files), and the `L191` in Changelog states that *"Remove parameters from `x/evidence` genesis and module state. The `x/evidence` module now solely uses Tendermint consensus parameters to determine of evidence is valid or not."* In fact, there seems to still remain a **problem (Inconsistency between design and implementation)** in the parameter setting because after PR5952, the setting in ComosHub adopts the default setting of [Tendermint](https://github.com/tendermint/tendermint/blob/35581cf54ec436b8c37fabb43fdaa3f48339a170/types/params.go#L47). But if we check the current-parameters.js in gaia, we can find the [max_age_num_blocks](https://github.com/cosmos/gaia/blob/c5de5d08816474c15a55a91c4a06d755805440a9/docs/docs/governance/current-parameters.js#L39) is **10 times than** the default value in Tendermint. Besides, the average block creation interval for Comos Hub is 6 seconds (which is different from the default value). So, the final validity period for evidence in gaia is about **MAX(70 days,  2 days)**, this is a strange situation because these two values are expected to be close.

I believe the security guarantee for Celestia should be close to that for Osmosis. This is because for Cosmos Hub, the exploit of malicious behaviour usually gains the native token, and the punishment mechanism will make it not profitable. However, the potential consensus instability risk for Osmosis or Celestia may lead to other losses. Overall, the genesis settings indicate that the maxAgeDuration equaling the unbonding time is a typical implementation.

cips/cip-draft_msrgpi_abbrev.md
caomingpei: Thanks, Updated.

cips/cip-draft_msrgpi_abbrev.md
rootulp: Thanks again for your thorough investigation! Agreed the Cosmos Hub situation seems complex. @mpoke is there any rationale for the inconsistency?

Given all that info, it seems simpler to set both evidnece.MaxAgeDuration  and unbonding time to the same values (21 days) like you initially proposed.

cips/cip-draft_msrgpi_abbrev.md
```diff
@@ -0,0 +1,124 @@
+---
```
jcstein, line 1:
```suggestion
---
cip: 16
```

jcstein (approved):
LGTM, cc @rootulp for a last review of relevant changes.

rootulp (approved):

jcstein (approved):

PR #76, opened 2024-02-08
Update cip-6.md title and add summary table

ebuchman:
In the spirit of CIP-13, added a single-row table summarizing the new parameter.

Would be good to set a standard to have this table in any CIP that introduces a new parameter or modifies one so it can easily be copied into CIP-13 if/when it's approved.

In this vein it wasn't entirely clear to me what the param store will be called. This CIP mentions it will be a new param.Subspace but doesn't name it, so I'm optimistically calling it `gas` but feel free to correct. That said it's not clear to me why this shouldn't just go under `auth` along with eg. `auth.TxSizeCostPerByte`? 

Also updated the title to be a bit more descriptive

## Checklist

### DIFF
--- a/cips/cip-6.md
+++ b/cips/cip-6.md
@@ -16,7 +16,7 @@ Implement a global, consensus-enforced minimum gas price on all transactions. En
 
 | Parameter     | Default | Summary                                                                                                                | Changeable via Governance |
 |---------------|---------|------------------------------------------------------------------------------------------------------------------------|---------------------------|
-| gas.MinimumGasPrice | 0.002utia  | Globally set minimum price per unit of gas                                                            | True                     |
+| minfee.MinimumGasPrice | 0.002utia  | Globally set minimum price per unit of gas                                                            | True                     |
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


cips/cip-6.md
```diff
@@ -14,13 +14,17 @@ created: 2023-11-30
 
 Implement a global, consensus-enforced minimum gas price on all transactions. Ensure that all transactions can be decoded and have a valid signer with sufficient balance to cover the cost of the gas allocated in the transaction. The minimum gas price can be modified via on-chain governance.
 
+| Parameter     | Default | Summary                                                                                                                | Changeable via Governance |
+|---------------|---------|------------------------------------------------------------------------------------------------------------------------|---------------------------|
+| gas.MinimumGasPrice | 0.002utia  | Globally set minimum price per unit of gas                                                            | True                     |
```
cmwaters, line 19:
```suggestion
| minfee.MinimumGasPrice | 0.002utia  | Globally set minimum price per unit of gas                                                            | True                     |
```

I think we were going to call the module `minfee`. Also it still seems like there's debate on what the initial/default should be but we can leave it at 0.002 for now

cmwaters (approved):
Thanks @ebuchman! Yeah I think a table makes it clearer

jcstein (approved):
LGTM if @cmwaters approves 🤝

PR #96, opened 2024-03-06
docs(CIP-14): define proposed param values

rootulp:
Currently CIP-14 is ambiguous with respect to the values it proposes for the host module parameters. The table in this PR aims to remove that ambiguity.

The values were chosen based on the latest messages in the [forum post](https://forum.celestia.org/t/moving-toward-safer-and-more-aligned-tia-liquid-staking/1422/18). I would appreciate reviews from the CIP authors: @womensrights @asalzmann @sampocs 

Note: I think we also need to fill out the **Test Cases** section of this CIP prior to moving it to review


jcstein:
Per CIP-1, test cases is optional, but for this CIP, I think it is helpful https://github.com/celestiaorg/CIPs/blob/main/cips/cip-template.md#test-cases

cips/cip-14.md
```diff
@@ -28,39 +28,48 @@ By enabling ICA on an accelerated timeline, Celestia can enable battle-tested pr
...
+| icahost.AllowMessages | `["/ibc.applications.transfer.v1.MsgTransfer", "/cosmos.bank.v1beta1.MsgSend", "/cosmos.staking.v1beta1.MsgDelegate", "/cosmos.staking.v1beta1.MsgBeginRedelegate", "/cosmos.staking.v1beta1.MsgUndelegate", "/cosmos.staking.v1beta1.MsgCancelUnbondingDelegation", "/cosmos.distribution.v1beta1.MsgSetWithdrawAddress", "/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward", "/cosmos.distribution.v1beta1.MsgFundCommunityPool", "/cosmos.gov.v1.MsgVote"]` | Provides the ability for a chain to limit the types of messages or transactions that hosted interchain accounts are authorized to execute by defining an allowlist using the Protobuf message type URL format. | False                     |
```
ebuchman, line 38: Thanks for the table!

> False

Should this be False? Potentially there may be desire to add more in the future?

cips/cip-14.md
```diff
@@ -28,39 +28,48 @@ By enabling ICA on an accelerated timeline, Celestia can enable battle-tested pr
...
+
+### Definitions
 
 - `Host Chain`: The chain where the interchain account is registered. The host chain listens for IBC packets from a controller chain which contain instructions (e.g. cosmos SDK messages) that the interchain account will execute.
```
ebuchman, line 42: Just realized a lot of this text is copied verbatim from the ICS27 spec in https://github.com/cosmos/ibc/tree/main/spec/app/ics-027-interchain-accounts . I think its good to have it here but maybe worth noting its a direct copy?

cips/cip-14.md
rootulp: I expect more messages to be added to this list in the future but IMO this list should be expanded via the CIP process rather than on-chain governance.

cips/cip-14.md
rootulp: can do!

jcstein: tagging authors @womensrights @asalzmann @sampocs for review

jcstein (approved):
LGTM, but will wait on review from @womensrights @asalzmann and @sampocs

womensrights:
lgtm, I will submit a seperate pr linking to test cases in the ibc-go repo

sampocs (approved):
lgtm, thanks for updating!

PR #109, opened 2024-03-20
cip-19: Review status

Wondertan:
Change status to review

### DIFF
--- a/cips/cip-19.md
+++ b/cips/cip-19.md
@@ -4,7 +4,7 @@ title: Shwap Protocol
 description: Shwap - a new messaging framework for DA and sampling
 author: Hlib Kanunnikov (@Wondertan)
 discussions-to: https://forum.celestia.org/t/cip-shwap-protocol/1551
-status: Draft
+status: Review
 type: Standards Track
 category: Data Availability, Networking
 created: 2024-02-02


Wondertan:
Or is it In Review? Also, where is the list of all the possible statuses?

jcstein:
Review is correct, in CIP-1, https://github.com/celestiaorg/CIPs/blob/main/cips/cip-1.md#cip-process

Wondertan:
Found it, CIP-1

jcstein:
While CIP-1 doesn't mention anything about how long a CIP remains in Draft before going to Review, should this remain in Draft longer than just a few hours?

Wondertan:
Actually, @jcstein, let's keep it as a draft; I will add one more container and ID type for the eds.

rootulp:
> While CIP-1 doesn't mention anything about how long a CIP remains in Draft before going to Review, should this remain in Draft longer than just a few hours? cc @YazzyYaz @rootulp @ebuchman

IMO there doesn't need to be a minimum time enforced on how long a CIP is in Draft before moving to Review.

> Actually, @jcstein, let's keep it as a draft; I will add one more container and ID type for the eds.

Sounds good. Should this PR be closed?

Wondertan:
We may, but the transition to review will happen anyway, and it's also the case that the change will be strictly additive and won't change the existing doc, so it still makes sense to keep what exist in the review stage

jcstein (approved):

Issue #112, opened 2024-03-21
Update CIP-6 to accurately reflect the current state of implementation


ninabarbakadze:
## Description

CIP-6 was implemented and merged view pr: https://github.com/celestiaorg/celestia-app/pull/3094/files#diff-008f85e59c1e0f88e9ad4be82fd1a15f29261f9b9d1a71a19488c58b94fef363R200

`cip-6.md` [reference implementation](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-6.md#reference-implementation) should be updated to reflect the changes.

ninabarbakadze:
yes the reference implementation section. I'll update the issue to reflect that. @rootulp

cmwaters:
Are we also able to move it into Final? I'll take one last pass to make sure everything is covered.

rootulp:
FYI @ninabarbakadze https://github.com/celestiaorg/CIPs/pull/121 merged but didn't close this issue. I think because the hyphen in the PR description.

PR #121, opened 2024-04-03
docs(cip-6): update the reference implementation

ninabarbakadze:
## Overview

Fixes - #112

### DIFF
--- a/cips/cip-6.md
+++ b/cips/cip-6.md
@@ -4,7 +4,7 @@ title: Mininum gas price enforcement
 description: Enforce payment of the gas for a transaction based on a governance modifiable global minimum gas price 
 author: Callum Waters (@cmwaters)
 discussions-to: https://forum.celestia.org/t/cip-006-price-enforcement/1351
-status: Last Call
+status: Final
 last-call-deadline: 2024-03-06
 type: Standards Track
 category: Core
--- a/cips/cip-6.md
+++ b/cips/cip-6.md
@@ -81,7 +81,7 @@ if err != nil {
 
 This will now reject undecodable transactions. Decodable transactions will still have to pass the `AnteHandlers` before being accepted in a block so no further change is required.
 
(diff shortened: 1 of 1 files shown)


jcstein:
let's cover move from last call to final in a new PR, as I think it is out of scope for the issue

ninabarbakadze:
> let's cover move from last call to final in a new PR, as I think it is out of scope for the issue

reverting the change

jcstein (approved):

cips/cip-6.md
```diff
@@ -81,9 +81,9 @@ if err != nil {
 
 This will now reject undecodable transactions. Decodable transactions will still have to pass the `AnteHandlers` before being accepted in a block so no further change is required.
 
-A mechanism to enforce a minimum fee is already in place in the `DeductFeeDecorator`. Currently, this is the validators locally set value sourced from their config. To switch to a network-wide value, we introduce a new `param.Subspace` which solely consisty of the min gas price. If this is unpopulated, it will be set to the default value of `0.002utia` (which matches the current local min gas price). As a `param.Subspace`, this value can be modified via governance as can any other parameter.
+A mechanism to enforce a minimum fee is already in place in the `DeductFeeDecorator`. Currently, this is the validators locally set value sourced from their config. To switch to a network-wide value, we introduce a new `param.Subspace`, `minfee`, which solely consists of the global min gas price. If this is unpopulated, it will be set to the default value of `0.002utia` (which matches the current local min gas price). As a `param.Subspace`, this value can be modified via governance as can any other parameter.
```
rootulp, line 84: > To switch to a network-wide value, we introduce a new `param.Subspace`, `minfee`, which solely consists of the global min gas price.

AFAIK the decorator isn't _switching_ but it's adding an additional rule. The old rule that uses the validators locally set config still applies. How about something like:

```suggestion
A mechanism to enforce a minimum fee is already in place in the `DeductFeeDecorator`. Currently, the decorator uses a  min gas price sourced from the validator's local config. To introduce a network-wide constraint on min gas price, we introduce a new `param.Subspace=minfee`, which contains the global min gas price. If the param is unpopulated, it defaults to `0.002utia` (which matches the current local min gas price).
```

ninabarbakadze: sounds good to me

rootulp (approved):

PR #132, opened 2024-04-17
chore: remove RIM from Lemongrass

rootulp:
## Motivation

My impression from today's call is that we're not including CIP-12 (Relayer Incentivization Middleware) in Lemongrass.


jcstein (approved):
Agreed. I think the auditing of CIP-12 would need to happen before it's added to Mainnet Beta. Due to this, best that this is removed to not block the Lemongrass Hardfork.

PR #137, opened 2024-04-23
Move cip-10 to final call and add a few more clarifying remarks

cmwaters:
## Overview

As well as the status update, this adds some more information on the upgrade process and provides a list of test cases.


rootulp (commented):
LGTM with one addition for a last call date

jcstein (approved):

PR #138, opened 2024-04-23
docs(cip-14): move to last call

rootulp:
CIP-14 hasn't had any significant changes recently. Same for the relevant [forum post](https://forum.celestia.org/t/moving-toward-safer-and-more-aligned-tia-liquid-staking/1422).

I propose moving it to last call with a deadline two weeks from now.


rootulp:
I added this to the agenda for next call.

jcstein (approved):

PR #147, opened 2024-05-22
CIP to introduce authored blobs

cmwaters:
This CIP is in reference to this issue: https://github.com/celestiaorg/celestia-app/issues/3316

Note that there is a diagram so when changing the name of the file and folders, make sure to change the name of the folder in the assets directory and update this CIP that references it


cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,92 @@
...
+For rollups, using ZK, such as the case with Soverign, the flow is as follows:
+
+- Query all the blobs from the rollup namespace via RPC
+- For each blob, reconstruct the blob commitment.
+- Fetch the PFB namespace
+- Parse the PFB namespace and create a mapping from blob commitment -> PFB
+- (In zero-knowledge) Accept the list of all rollup blobs and the list of relevant PFBs as inputs
+- (In zero-knowledge) Verify that the claimed list of blobs matches the block header using the namespaced merkle tree
+- (In zero-knowledge) For each blob, find the PFB with the matching commitment and check that the sender is correct.
+- (In zero-knowledge) For each relevant PFB, check that the bytes provided match the namespaced merkle tree
```
evan-forbes, line 33: > (In zero-knowledge)

From my perspective, I think I know what is meant by "zero-knowledge", but it might be valuable to be more specific or use different wording. I've heard the term "in circuit" used before for constraints applied to the public and private inputs, do you think that applies here or perhaps something else?

evan-forbes (commented):
still chewing but otherwise LGTM!

cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,92 @@
...
+discussions-to: https://forum.celestia.org/t/cip-blobs-with-verified-author
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-05-22
+---
+
+## Abstract
+
+Introduce a new v2 blob type that can be submitted with the author of the blob. Validators verify that the author is correct, simplifying the loop for rollups that adopt a fork-choice rule that enshrines a specific sequencer.
```
preston-evans98, line 14: nit: This is useful for rollups that don't enshrine a sequencer as well! In fact, we don't enshrine a particular sequencer in Sovereign SDK

cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,92 @@
...
+
+Given proto's backwards compatibility, users could still submit the old blob type (in the `BlobTx` format) and signer would be processed as an empty string.
+
+The new block validity rule (In `PrepareProposal` and `ProcessProposal`) would thus be that if the signer was not empty, then it must match that of the PFB that paid for it. When validating the `BlobTx`, validators would check the equivalency of the PFB's `signer` to the Blob's `signer` (as well as verification of the signature itself).
+
+Although no version changes are required for protobuf encoded blobs, share encoding would change. Blobs containing a non empty signer string would be encoded using the new v2 format:
+
+![Diagram of V2 Share Format](../assets/cip-blob-verified-signer/blob-v2-share-format.svg)
+
+Blobs with an empty `signer` string would remain encoded using the v1 format. Note that in this diagram it is the `Info Byte` that contains the share version. Not to be confused with the namespace version.
```
preston-evans98, line 60: If the `signer` is empty but a user specifies `share_version = 2`, is the version in the `Info Byte` also changed to be `v1`?

cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,92 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+## Rationale
+
+Given the current specification change, the new loop is simplified:
+
+- Retrieve all blobs in the subscribed namespaces
+- Verify the namespaces inclusion
+- Verify that the `signer` in each blob matches that of the sequencer
```
preston-evans98, line 70:
```suggestion
- Verify that the `signer` in each blob matches that of an allowed sequencer
```

cips/cip-blob-verified-signer.md
cmwaters: Can you explain that a little further. My mental model was that the nodes had a list of accepted "authors" of blobs. Or perhaps another way of putting it is If anyone can submit blobs then why do you need to verify the signer.

cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,92 @@
...
+
+As a small digression, it may be feasible to additionally introduce a new namespace version with the enforcement that all blobs in that namespace use the v2 format i.e. have a signer. However, this does not mean that the signer matches that of the sequencer (which Celestia validators would not be aware of). This would mean that full nodes would need to get and verify all blobs in the namespace anyway.
```
preston-evans98, line 72: This is a good call out. If Celestia also added an RPC endpoint for fetching blob metadata, this could be made quite efficient. Rollups would just have to...
1. Fetch the metadata via RPC
2. For each blob from an allowed sender, fetch the full blob and its proof via RPC
3. (zk-rollups only) For each other blob, fetch *only the first share* of the blob and its proof via RPC to check that the metadata is correct

cips/cip-blob-verified-signer.md
cmwaters: I was taking this from a slack thread. I can rephrase to using (in circuit) if you think that would be better

cips/cip-blob-verified-signer.md
cmwaters: As I see it (and feel free to correct me if you feel there's a better configuration), a tx specifying a share version of 2 and an empty signer field will be rejected. A tx specifying a share version of 1 with a signer field will also be rejected.

cips/cip-blob-verified-signer.md
cmwaters: Iiuc, you mean that there would be a RPC query to fetch blobs by namespace and signer?

cips/cip-blob-verified-signer.md
preston-evans98: We have a list of accepted authors, but it's dynamic (no particular enshrined sequencer), and we even accept blobs from people who aren't on the list under some circumstances (i.e. to enable recovery if all sequences are offline/malicious)

preston-evans98: The actual rules are a little bit involved, but the high level idea is that we accept...
- All blobs from registered sequencers (i.e. actors who are subject to slashing on the rollup)
- Up to a (small) constant number of additional blobs per celestia block from unregistered sequencers

cips/cip-blob-verified-signer.md
preston-evans98: That sounds great. Can we add a sentence making it explicit in the CIP?

cips/cip-blob-verified-signer.md
cmwaters: I see so it's somewhat a middle ground between complete permissionless rollups and enshrined (whitelisted). I'll update the abstract accordingly

cips/cip-blob-verified-signer.md
cmwaters: So another way of framing this would be the minimal set of data for the fork-choice rule to filter out what blobs are worth processing or not. Wouldn't this data need to be provable? Wouldn't this make it look similar to the `PFBNamespace` although a little more compact

nashqueue:
Maybe we should add the fact that you need versioning for the CreateCommitment function. As we might be adding other things in the future which will change the shares and therefore the share commitment. 

Create commitment will now have to be aware of the signer.

cips/cip-blob-verified-signer.md
cemozerr: > Up to a (small) constant number of additional blobs per celestia block from unregistered sequencers

The constant is so small that effectively rollup should be completely permissionless.

cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,92 @@
...
+Blobs with an empty `signer` string would remain encoded using the v1 format. Note that in this diagram it is the `Info Byte` that contains the share version. Not to be confused with the namespace version.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+## Rationale
+
+Given the current specification change, the new loop is simplified:
+
+- Retrieve all blobs in the subscribed namespaces
+- Verify the namespaces inclusion
```
rootulp, line 69: [nit] can this be clarified b/c I'm not sure what it means. Does this step mean:

- Verify that the blobs from the previous step are present in the subscribed namespace?
- Verify that the blobs from the previous step are present and complete in the subscribed namespace? In other words, there are no other blobs in the subscribed namespace.
- Verify the blobs from the previous step are included in an NMT with root `RowRoot`? Additionally verify a `RowRoot` is included in some `DataRoot`?

cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,92 @@
...
+
+Test cases will need to ensure that a user may not forge a incorrect signer, nor misuse the versioning. Test cases should also ensure that the properties of backwards compatibility mentioned earlier are met.
+
+## Reference Implementation
+
+TBC
+
+## Security Considerations
+
+Rollups using this pattern for verifying the enshrined sequencer make an assumption that there is at least 1/3 in voting power of the network is "correct". Note this is a more secure assumption than forking which may require up to 2/3+ of the voting power to be "correct". Rollups may decide to still retrieve the PFB's and validate the signatures themselves if they wish to avoid this assumption.
```
rootulp, line 88: I don't follow this security consideration. 

> Note this is a more secure assumption than forking which may require up to 2/3+ of the voting power to be "correct".

Does "forking" in this context apply to Celestia or the roll-up? Why do 2/3 of the voting power need to be correct to fork Celestia? Hypothetically if 2/3 voting power on the Celestia consensus layer are malicious, the 1/3 could create a fork that excludes the 2/3 malicious nodes.

> Rollups may decide to still retrieve the PFB's and validate the signatures themselves if they wish to avoid this assumption.

I think this is trying to say that regardless of the number of "incorrect" voting power on the DA layer, the DA consensus layer can't include a PFB with an signature that would convince an honest roll-up node to be deceived. But I don't really understand how verifying the PFB signatures avoids an assumption from above about 1/3 or 2/3 correct voting power.

cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,92 @@
+---
+title: Support authored blobs
```
rootulp, line 2: [optional] this title is a bit unclear because blobs are already authored by blob submitters. An alternative title that may be a bit more self-evident is: "Include signer in share format"

rootulp (commented):
Thanks for including the diagram!

cips/cip-blob-verified-signer.md
cmwaters: But blobs don't have the author / signer in the blob itself. It's embedded in the PFB. And this isn't just about changes to the share format

cmwaters:
> Maybe we should add the fact that you need versioning for the CreateCommitment function. As we might be adding other things in the future which will change the shares and therefore the share commitment.
> 
> Create commitment will now have to be aware of the signer.

The CreateCommitment function has the following signature:

```
func CreateCommitment(blob *blob.Blob, merkleRootFn MerkleRootFn, subtreeRootThreshold int) ([]byte, error) 
```

As it includes the blob, which has the signer as well as the share version, I don't think anything will need to change to the function signature. Of course within the function, the commitment construction will depend on whether the signer is present in the blob or not

cips/cip-blob-verified-signer.md
cmwaters: > Does "forking" in this context apply to Celestia or the roll-up? 

To Celestia (which also implies that there is a potential fork of all the rollups on Celestia).

What I was saying is that there is an assumption on the honesty of the validator set when using the signer (since the validators validate that the signer is correct) but that this assumption is in any way outweighed by the bigger assumption (i.e. the less secure attack vector) that there isn't currently a fork on Celestia

cips/cip-blob-verified-signer.md
cmwaters: I would assume that inclusion and exclusion are done simultaneously i.e. we are proving that all the blobs we have are present in that namespace and there are no blobs in that namespace that we don't have.

Does that make sense? And would you like me to elaborate more in this document

cips/cip-blob-verified-signer.md
preston-evans98: This data is already provable, since it's encoded in the start share of each blob anyway. But TBC, we at Soveregin don't actually need/want this - I was just pointing it out in case it's useful for optimistic rollups

cips/cip-blob-verified-signer.md
rootulp: That sounds like the second option:

> Verify that the blobs from the previous step are present and complete in the subscribed namespace? In other words, there are no other blobs in the subscribed namespace.

But I think this would be clearer than what I previously wrote:

```
1. Retrieve all blobs in the subscribed namespace
2. Verify that the blobs from the previous step belong to the subscribed namespace and the list is complete (i.e. there are no blobs present in the subscribed namespace that are absent from the list of retrieved blobs).
```

cips/cip-blob-verified-signer.md
rootulp: Ahh got it, thanks for expanding here.

cips/cip-blob-verified-signer.md
rootulp: Good point. An alternative name could be: "Blob v2 with signer"

cips/cip-blob-verified-signer.md
```diff
@@ -0,0 +1,94 @@
...
+Blobs with an empty `signer` string would remain encoded using the v1 format. A transaction specifying a share version of 2 and an empty signer field would be rejected. Equally so, specifying a share version of 1 and a non empty signer field would be rejected.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+## Rationale
+
+Given the current specification change, the new loop is simplified:
+
+- Retrieve all blobs in the subscribed namespaces
+- Verify the blobs inclusion by namespace
```
mastergaurang94, line 71: (Disclaimer: I'm a newb) what do you mean? Aren't the blobs returned guaranteed to be included?

cips/cip-blob-verified-signer.md
```diff
@@ -57,7 +57,9 @@ Although no version changes are required for protobuf encoded blobs, share encod
 
 ![Diagram of V2 Share Format](../assets/cip-blob-verified-signer/blob-v2-share-format.svg)
 
-Blobs with an empty `signer` string would remain encoded using the v1 format. Note that in this diagram it is the `Info Byte` that contains the share version. Not to be confused with the namespace version.
+Note that in this diagram it is the `Info Byte` that contains the share version. Not to be confused with the namespace version.
+
+Blobs with an empty `signer` string would remain encoded using the v1 format. A transaction specifying a share version of 2 and an empty signer field would be rejected. Equally so, specifying a share version of 1 and a non empty signer field would be rejected.
```
rootulp, line 62: [nit] the current share version is `0` so a new share version would be `1`. To avoid confusion, we may want to replace v1 => 0 and v2 => 1 in these lines.

Ref: https://github.com/celestiaorg/celestia-app/blob/c5b0b9421e8bbe04d26104716bb86750a6891ab0/specs/src/specs/shares.md?plain=1#L49-L50

rootulp (approved):

cips/cip-blob-verified-signer.md
jcstein: let's do this and I think we can get this (1) assigned a CIP number (2) edit the image directory and path in .md (3) and merged

cips/cip-blob-verified-signer.md
jcstein: what do you think is best here @evan-forbes and @cmwaters, "in circuit" or "in zero-knowledge"?

cips/cip-blob-verified-signer.md
jcstein: is this resolved?

cips/cip-blob-verified-signer.md
jcstein: is this resolved?

cips/cip-blob-verified-signer.md
jcstein: > `The location of the blob. Alternatively, we could put the blob commitment`
I think that this is not going to be the most atomic way to point to a blob. If I submit 2 blobs that are identical, their commitments will be the same, and their index will also be different, although they have the same commitment.

cips/cip-blob-verified-signer.md
jcstein: is this resolved?

jcstein (changes requested):
LGTM, but would like to resolve the open discussions before merging

cips/cip-blob-verified-signer.md
rootulp: Yea resolved.

cips/cip-blob-verified-signer.md
cmwaters: Yes this is resolved

cips/cip-blob-verified-signer.md
cmwaters: Yes ok, I can also refer to them as v0 and v1

cips/cip-blob-verified-signer.md
cmwaters: Resolved: https://github.com/celestiaorg/CIPs/pull/147/commits/62e5ef1ac8ffb785e5777cd837439d24b8a201f6

cips/cip-blob-verified-signer.md
cmwaters: Changed to in circuit (https://github.com/celestiaorg/CIPs/pull/147/commits/62e5ef1ac8ffb785e5777cd837439d24b8a201f6)

cips/cip-blob-verified-signer.md
cmwaters: Resolved in https://github.com/celestiaorg/CIPs/pull/147/commits/62e5ef1ac8ffb785e5777cd837439d24b8a201f6

cmwaters:
@jcstein all comments should be resolved now

cips/cip-blob-verified-signer.md
```diff
@@ -53,25 +53,25 @@ Given proto's backwards compatibility, users could still submit the old blob typ
 
 The new block validity rule (In `PrepareProposal` and `ProcessProposal`) would thus be that if the signer was not empty, then it must match that of the PFB that paid for it. When validating the `BlobTx`, validators would check the equivalency of the PFB's `signer` to the Blob's `signer` (as well as verification of the signature itself).
 
-Although no version changes are required for protobuf encoded blobs, share encoding would change. Blobs containing a non empty signer string would be encoded using the new v2 format:
+Although no version changes are required for protobuf encoded blobs, share encoding would change. Blobs containing a non empty signer string would be encoded using the new v1 share format (the first share format version is 0):
 
-![Diagram of V2 Share Format](../assets/cip-blob-verified-signer/blob-v2-share-format.svg)
+![Diagram of V1 Share Format](../assets/cip-blob-verified-signer/blob-v2-share-format.svg)
```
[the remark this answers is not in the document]
cmwaters: Yeah could the CIP maintainers update the file name and reference when they assign the CIP number

rootulp:
Assigned this CIP-21. FYI @jcstein

cips/cip-blob-verified-signer.md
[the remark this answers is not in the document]
rootulp: Done in https://github.com/celestiaorg/CIPs/pull/147/commits/0876f5bef4e04aad5770f2102fae29b8f2d06181

cips/cip-21.md
```diff
@@ -0,0 +1,94 @@
+---
```
jcstein, line 1:
```suggestion
---
cip: 21
```

jcstein (approved):

PR #159, opened 2024-06-20
cip-19!: simplify Bitswap compostion

Wondertan:
While the status of the CIP is still `Review`, we want to make a protocol-breaking amendment to Shwap prompted by the ongoing production-ready implementation. This change simplifies the protocol, particularly the Bitswap composition, by making a single protobuf definition of a Block instead of N definitions per every current and future Container type.

This change is motivated by a desire to make it easier for Shwap to extend horizontally in the abstract specification and its implementations. Previously, the new container would have to be added to both the core Shwap container specification and the Bitswap composition. Now, only a single place needs to be updated(and Bitswap codec registry). 

Besides, decoupling Shwap's and composition's protobuf definitions significantly improved cohesion across the layers.


Wondertan:
(Noting that we should not rush moving Shwap to `Final`, as we figured there is one more feature needs to be added -- ability to request only Container proofs without its data)

jcstein (approved):

PR #161, opened 2024-07-01
move CIP-21: Authored Blobs into review

cmwaters:


jcstein (approved):

rootulp (approved):
The reference implementation section needs to be filled out but I don't think that strictly needs to block moving this to review

cmwaters:
> The reference implementation section needs to be filled out but I don't think that strictly needs to block moving this to review

True. I think the specification should be sufficient. I was planning on writing up the reference implementation once we have one. I'll aim to get that done before this goes into final call

YazzyYaz (approved):

PR #164, opened 2024-07-03
CIP: Coordinated Start Time Intervals

cmwaters:
This CIP outlines protocol changes to support more consistent average block times regardless of the size of the blocks (to a point).

PR #165, opened 2024-07-12
CIP: Coordinated prevote times

cmwaters:
Supercedes: https://github.com/celestiaorg/CIPs/pull/164

### DIFF
--- a/cips/cip-prevote.md
+++ b/cips/cip-prevote.md
@@ -2,7 +2,7 @@
 title: Coordinated prevote times
 description: Scheduled prevote times for consistent blocks
 author: Callum Waters (@cmwaters)
-discussions-to: URL
+discussions-to: https://forum.celestia.org/t/coordinated-start-time-intervals/1768
 status: Draft
 type: Standards Track
 category: Core
--- a/cips/cip-prevote.md
+++ b/cips/cip-prevote.md
@@ -1,0 +1,72 @@
+---
+title: Coordinated prevote times
+description: Scheduled prevote times for consistent blocks
+author: Callum Waters (@cmwaters)
+discussions-to: URL
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-07-12
+---
+
+## Abstract
+
+Block propagation speed is dependent on block size, thus larger blocks have later block times. It is desirable that regardless of block size that blocks are finalized with a consistent cadence. This CIP proposes a solution to this problem, namely, correct processes that receive a valid proposal withthold their prevote until the `ProposalTimeout` has elapsed.
+
+## Motivation
+
+The current network has a cap of 2MB. We have observed in testnets that larger blocks (~8MB) shift the block time by up to 7 seconds (18 . seconds in total with the default 11 second `TimeoutCommit`). This range of 1 - 7 seconds for blocks 0-8MB in size is too inconsistent for users. We want to provide block finality with a consistent cadence.
+
+## Specification
+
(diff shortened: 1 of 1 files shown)


cips/cip-prevote.md
```diff
@@ -0,0 +1,72 @@
...
+created: 2024-07-12
+---
+
+## Abstract
+
+Block propagation speed is dependent on block size, thus larger blocks have later block times. It is desirable that regardless of block size that blocks are finalized with a consistent cadence. This CIP proposes a solution to this problem, namely, correct processes that receive a valid proposal withthold their prevote until the `ProposalTimeout` has elapsed.
+
+## Motivation
+
+The current network has a cap of 2MB. We have observed in testnets that larger blocks (~8MB) shift the block time by up to 7 seconds (18 . seconds in total with the default 11 second `TimeoutCommit`). This range of 1 - 7 seconds for blocks 0-8MB in size is too inconsistent for users. We want to provide block finality with a consistent cadence.
```
rootulp, line 18: [nit] errant period.

nit: does it make sense to rephrase the range in terms of the observed block interval because that is what users perceive: "This range of 1 - 7 seconds" => "This range of 11 - 18 seconds"

```suggestion
The current network has a cap of 2MB. We have observed in testnets that larger blocks (~8MB) shift the block time by up to 7 seconds (18 seconds in total with the default 11 second `TimeoutCommit`). This range of 11 - 18 seconds for blocks 0-8MB in size is too inconsistent for users. We want to provide block finality with a consistent cadence.
```

cips/cip-prevote.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Specification
+
+The current system relied on a fixed timeout (known as `TimeoutCommit`) after finality to dictate the interval between block times. The proposed modification targets the proposal timeout. Currently, if a correct process does not receive a proposal within a set time `TimeoutPropose` after the rounds start time, the process will prevote nil. This mechanism ensures livenss in the event that the proposer fails. This CIP makes the following modification:
```
rootulp, line 22:
```suggestion
The current system relied on a fixed timeout (known as `TimeoutCommit`) after finality to dictate the interval between block times. The proposed modification targets the proposal timeout. Currently, if a correct process does not receive a proposal within a set time `TimeoutPropose` after the round's start time, the process will prevote nil. This mechanism ensures liveness in the event that the proposer fails. This CIP makes the following modification:
```

rootulp (approved):
LGTM with nits. I think this can merge after a forum post URL is added

cips/cip-prevote.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Coordinated prevote times
+description: Scheduled prevote times for consistent blocks
+author: Callum Waters (@cmwaters)
+discussions-to: URL
```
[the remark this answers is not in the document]
cmwaters: Should I create a new one or just morph the previous. The problem is still the same, this just proposes a different solution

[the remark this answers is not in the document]
rootulp: Seems fine to use the previous forum link https://forum.celestia.org/t/coordinated-start-time-intervals/1768/6 b/c there's already some good discussion there on this problem.

rootulp, line 5:
```suggestion
discussions-to: https://forum.celestia.org/t/coordinated-start-time-intervals/1768
```

rootulp (approved):

PR #167, opened 2024-07-15
[DRAFT] canonical v1 API

liamsi:


assets/cip-api-v1/proto/blob/blob.proto
```diff
@@ -0,0 +1,76 @@
...
+message Blob {
+  bytes namespace_id = 1;
+  bytes data = 2;
+  uint32 share_version = 3;
+  bytes commitment = 4;
+  uint32 index = 5;
+  // TODO do we need namespace version in here as well?
+}
+
+message BlobRequest {
```
liamsi, line 18: https://github.com/celestiaorg/celestia-node/discussions/3517#discussioncomment-10047397

PR #197, opened 2024-08-20
docs(cip-19): add NamespaceRangeID along with NamespaceRangeContainer

vgonkivs:
## Overview
Added NamespaceRangeID that forms a request to get a range of namespace shares along with their inclusion proofs. The shares MAY span across multiple rows in the DataSquare.


cips/cip-19.md
```diff
@@ -311,6 +311,49 @@ Namespace data may span over multiple rows, in which case all the data is encaps
 containers. This enables parallelization of namespace data retrieval and certain [compositions](#protocol-compositions)
 may get advantage of that by requesting containers of a single namespace from multiple servers simultaneously.
 
+#### NamespaceRangeID
```
zvolin, line 314: namespace range may suggest it's a range of namespaces, like 0x00000001 up to 0x00000004. Maybe just NamespaceData would be better? just dropping Row prefix from RowNamespaceData, also better showing correlation between the two?

cips/cip-19.md
```diff
@@ -311,6 +311,49 @@ Namespace data may span over multiple rows, in which case all the data is encaps
...
+
+NamespaceRangeID: encapsulates SampleID, Namespace and identifies the continuous range of shares in the DataSquare. Formated as below:
+
+```text
+NamespaceRangeID {
+  SampleID;
```
zvolin, line 320: what is the purpose of sample here? does it mean take `Length` shares from given `SampleID`? I think it needs better clarification below. Also why not just using `EdsId` + `Namespace` providing whole namespace from given block? Is there a usecase for a user to request his shares but eg. skipping the first and last one?

cips/cip-19.md
```diff
@@ -311,6 +311,49 @@ Namespace data may span over multiple rows, in which case all the data is encaps
...
+NamespaceRangeContainer containers encapsulate user-submitted data under [namespaces][ns] within a single or multiple
+of [DataSquare][square] rows. It MAY contain [shares][shares] and [NMT][nmt] proof of share inclusion or proof only.
+
+NamespaceRangeContainer container are protobuf formatted using the following proto3 schema:
+
+```protobuf
+syntax = "proto3";
+
+message NamespaceRangeContainer {
+  repeated RowNamespaceData namespacedData = 1;
```
zvolin, line 351:
```suggestion
  repeated RowNamespaceData namespaceData = 1;
```

makes it more consistent with other definitions

cips/cip-19.md
```diff
@@ -311,6 +311,49 @@ Namespace data may span over multiple rows, in which case all the data is encaps
...
+#### NamespaceRangeID
+
+NamespaceRangeID: encapsulates SampleID, Namespace and identifies the continuous range of shares in the DataSquare. Formated as below:
+
+```text
+NamespaceRangeID {
+  SampleID;
+  Namespace;
+  Length: u16;
+  OmitData: bool;
```
zvolin, line 323: should OmitData also be present in `RowNamespaceId`?

jcstein:
Will let @vgonkivs address comments from @zvolin and then give review 🫡

cips/cip-19.md
```diff
@@ -311,6 +311,49 @@ Namespace data may span over multiple rows, in which case all the data is encaps
...
+
+NamespaceRangeContainer containers encapsulate user-submitted data under [namespaces][ns] within a single or multiple
+of [DataSquare][square] rows. It MAY contain [shares][shares] and [NMT][nmt] proof of share inclusion or proof only.
+
+NamespaceRangeContainer container are protobuf formatted using the following proto3 schema:
+
+```protobuf
+syntax = "proto3";
+
+message NamespaceRangeContainer {
```
zvolin, line 350:
```suggestion
message NamespaceRange {
```
the container suffix should be dropped for consistency

cips/cip-19.md
Wondertan: It may also be part of SampleID, i.e., one can request just the proof for the share/sample without the data itself. However, we are unaware of any use cases for the Row and Sample containers that can be requested without data. We could add that to be consistent with OmitData usage across types, but our approach is to specify and implement only what's known to be helpful in practice.

cips/cip-19.md
vgonkivs: > does it mean take Length shares from given SampleID?

Yes

vgonkivs: > Also why not just using EdsId + Namespace providing whole namespace from given block? Is there a usecase for a user to request his shares but eg. skipping the first and last one?

This means we will retrieve the whole eds from the requested namespace. For example, `blob.GetProof` needs a specific range of shares rather than the whole namespace.

cips/cip-19.md
Wondertan: We have:
`RowNamespaceData` - namespace data within a row
`NamespaceData` - all the namespace data that can span over multiple rows

To avoid namespace confusion I propose to do:
`NamespaceDataRange` - namespace data range that can span over multiple rows

vgonkivs: Agree with @Wondertan. The container itself and request namings should contain `Range`

rootulp (commented):
LGTM after one last rename

jcstein (approved):
LGTM after @rootulp 's comment is committed, and after review from previous reviewers

cips/cip-19.md
zvolin: should we then specify here what should happen if cid is pointing to shares outside of it's namespace? eg. what if the sample id is not in correct namespace or the length goes beyond the namespace?

zvolin: also I'm not exactly sure what is the usecase here, user would need to already know the blob's position in eds and it's length in shares to even use that, however shwap itself doesn't give any of such abilities

vgonkivs:
Moving this PR to draft until we have a working prototype to avoid any changes in the future.

jcstein:
Please use a filename such as cip-add-hyperlane.md instead of assigning a number. See the example: https://cips.celestia.org/cip-template.html

PR #209, opened 2024-09-23
move cip 21 into last call

cmwaters:


cips/cip-21.md
```diff
@@ -4,7 +4,7 @@
 | description | Introduce a new blob type that can be submitted whereby the signer address is included and verified. |
 | author | Callum Waters ([@cmwaters](https://github.com/cmwaters)) |
 | discussions-to | <https://forum.celestia.org/t/cip-blobs-with-verified-author> |
-| status | Review |
+| status | Last Call |
```
rootulp, line 7: Needs a last call date

```suggestion
| status | Last Call |
| last-call-deadline | 2024-10-07 |
```

rootulp (commented):
The reference implementation still needs to be filled out

jcstein (dismissed):
- [x] Add reference implementation
- [x] Add last call deadline
+1 to @rootulp's review

cips/cip-21.md
```diff
@@ -84,7 +84,42 @@ Test cases will need to ensure that a user may not forge a incorrect signer, nor
...
-TBC
+The go implementation makes the following modifications to the `Blob` proto:
+
+```diff
+message BlobProto {
+  bytes namespace_id = 1;
+  bytes data = 2;
+  uint32 share_version = 3;
+  uint32 namespace_version = 4;
++ // Signer is sdk.AccAddress that paid for this blob. This field is optional
```
rootulp, line 95: [question] can this feature be used with feegrant? If yes, and we have:

1. celestiaA agrees to pay gas for all message for celestiaB 
2. celestiaB submits an authored blob and uses gas from celestiaA. 
3. Whose address goes in the signer field?

If it's celestiaB then should we stop using the term "that paid for this blob" and instead use something like: 

```go
// Signer is sdk.AccAddress that submitted this blob.
```

jcstein: good catch!

cips/cip-21.md
```diff
@@ -83,7 +84,42 @@ Test cases will need to ensure that a user may not forge a incorrect signer, nor
...
```
jcstein, line 95:
```suggestion
+ // Signer is sdk.AccAddress that signed this blob. This field is optional
```

rootulp:
Callum is OOO, maybe we should merge this as-is and we can edit it after the fact if that comment about "paid for" vs "submitted" is confusing for anyone else.

cc: @jcstein thoughts?

jcstein:
I think we can implement this feedback after it's in Last Call. However, here's some food for thought:

IIRC, the fee Granter/Payer account does not show any messages when blobs are posted. So If I had blob X with Grantee/Signer A and Granter/Payer B, it would only show on chain for Grantee/Signer A

| Account | Tx | Message on explorer? |
|---------|---|------------------------|
| A: Grantee/Signer | [PayForBlobs](https://mocha.celenium.io/tx/f28aeffeb003a21d4810660bb7d529b744d24419440e29080b3284a247977fe1?tab=messages) | Yes |
| B: Granter/Payer | PayForBlobs from "A: Grantee/Signer" | No, see [transactions](https://mocha.celenium.io/address/celestia1sung806mz9r4lwdpaw6usk7hcw42ku7kyun44w?tab=transactions) and [messages](https://mocha.celenium.io/address/celestia1sung806mz9r4lwdpaw6usk7hcw42ku7kyun44w?tab=messages) tabs |
| B: Granter/Payer | [FeeGrant](https://mocha.celenium.io/tx/802a17777fbeab416f6fa2c25f0c56dd9cc8a92afc2a96293d114ac7c22efb5c?tab=messages)  | Yes |

rootulp:
Thanks for the links! That seems like expected behavior because the fee granter didn't submit a `MsgPayForBlobs` in this example. So even though the fee granter paid for the blob to be included on-chain, the `MsgPayForBlobs` submitter (`celestia1rgj3z5rfydj3xulhm6glqr0zs7j8rr9rfh6ekw`) is the authored blob "signer"

cips/cip-21.md
```diff
@@ -83,7 +84,42 @@ Test cases will need to ensure that a user may not forge a incorrect signer, nor
...
+    if appVersion < v3.Version {
+      return ErrUnsupportedShareVersion.Wrapf("share version %d is not supported in %d. Supported from v3 onwards", blob.ShareVersion(), appVersion)
+    }
+    if !bytes.Equal(blob.Signer(), signer) {
+      return ErrInvalidBlobSigner.Wrapf("blob signer %s does not match msgPFB signer %s", sdk.AccAddress(blob.Signer()).String(), msgPFB.Signer)
+    }
+  }
+}
+```
+
```
rootulp, line 122: markdownlint is failing b/c too many empty lines

```suggestion
```

cips/cip-21.md
jcstein: > So even though the fee granter paid for the blob to be included on-chain, the MsgPayForBlobs submitter (celestia1rgj3z5rfydj3xulhm6glqr0zs7j8rr9rfh6ekw) is the authored blob "signer".

So in this case shouldn't it be `// Signer is sdk.AccAddress that signed this blob`?

cips/cip-21.md
jcstein: After reading through this CIP again recently, I'm a little confused how this would handle blobs submitted with feegrant as described in this thread.

PR #210, opened 2024-10-01
docs(cip-24): move to last call

rootulp:
We want to include this in v3 hence the shorter last call deadline of ~1 week. We'd like to get v3 on Arabica by Oct 7.


jcstein (approved):
LGTM

PR #216, opened 2024-10-10
feat: Versioned timeouts CIP draft

jcstein:

PR #217, opened 2024-10-10
feat: add six second CIP and compute CIP to ginger draft

jcstein:
## Overview

Updates Ginger Draft from suggestion by @rootulp in https://github.com/celestiaorg/CIPs/pull/216#discussion_r1795485634

Adds CIP-10 to Meta CIP for Ginger network upgrade

### DIFF
--- a/cips/cip-25.md
+++ b/cips/cip-25.md
@@ -7,7 +7,7 @@
 | status | Draft |
 | type | Meta |
 | created | 2024-10-01 |
-| requires | CIP-21, CIP-24, CIP-TBD |
+| requires | CIP-10, CIP-21, CIP-24, CIP-TBD |
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


cips/cip-25.md
```diff
@@ -17,9 +17,10 @@ This Meta CIP lists the CIPs included in the Ginger network upgrade.
 
 ### Included CIPs
 
+- [CIP-10](./cip-10.md): Coordinated network upgrades
```
rootulp, line 20: CIP-10 was already included in Lemongrass so IMO shouldn't be listed here. It's fine to list as a prerequisite at the top (next to requires) if you think it's needed



```suggestion
```

cips/cip-25.md
```diff
@@ -17,9 +17,10 @@ This Meta CIP lists the CIPs included in the Ginger network upgrade.
...
 - [CIP-21](./cip-21.md): Introduce blob type with verified signer
 - [CIP-24](./cip-24.md): Versioned Gas Scheduler Variables
-- CIP-TBD: limiting compute per block
+- CIP-TBD: Reduce block time to six seconds
```
rootulp, line 23: AFAIK we may have two CIPs still:
1. limiting compute per block
2. reducing block time to six seconds

cips/cip-25.md
jcstein: added this but accidentally didn't use the suggestion, my b!

rootulp (approved):

PR #226, opened 2024-10-17
docs: cip for tx limits

jcstein:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/3970


cips/cip-tx-limits.md
```diff
@@ -0,0 +1,34 @@
...
+
+This CIP proposes to set the limit for transaction size. The proposal is to set the transaction size limit to 2MiB. Setting the transaction size limit is consensus-breaking.
+
+## Specification
+
+1. Transaction size is limited to 2MiB by setting the versioned parameter `MaxTxSize` to 2097152, which is 2MiB in bytes. From version v3 and above, in `CheckTx`, `PrepareProposal`, and `ProcessProposal`, each transaction's size is checked against the `appconsts.MaxTxSize` threshold. This ensures that transactions over the limit are rejected or excluded at all stages, from initial submission to execution.
+
+## Rationale
+
+The rationale for this proposal is to set the transaction size limit to 2MiB, even with 8MiB blocks, to prevent issues with gossiping large transactions. Gossiping an 8MiB transaction without chunking could be detrimental to the network. This is a consensus-breaking change.
```
rootulp, line 22: @ninabarbakadze can you elaborate here on why we implemented this as a consensus breaking change instead of modifying the default mempool `MaxTxBytes` to 2 MiB?

I asked about it in a few places (Slack) and https://github.com/celestiaorg/celestia-app/issues/3686#issuecomment-2393738597

rootulp (approved):
Note this PR closes https://github.com/celestiaorg/celestia-app/issues/3970

Thanks for putting this up. I think it's safe to merge but it would be nice if @ninabarbakadze added more details before we move this from draft.

cips/cip-tx-limits.md
jcstein: I'll wait on feedback until tomorrow and plan to get draft in then. Thank you @ninabarbakadze and @rootulp !

ninabarbakadze: because every node can set the default to any value they choose to and afaik we don't know the effect large transactions are going to have on the network so in my understanding it's preventative. cc @evan-forbes

cips/cip-28.md
```diff
@@ -0,0 +1,34 @@
...
+
+The rationale for this proposal is to set the transaction size limit to 2MiB, even with 8MiB blocks, to prevent issues with gossiping large transactions. Gossiping an 8MiB transaction without chunking could be detrimental to the network. This is a consensus-breaking change.
+
+## Backwards Compatibility
+
+This proposal is meant to be included with v3 and the [Ginger Network Upgrade](./cip-25.md). It is a consensus-breaking change.
+
+## Security Considerations
+
+This proposal does not introduce any new security risks. However, it does impact network behavior and user experience, which should be carefully considered during implementation.
```
ninabarbakadze, line 30: Any changes to the block validity rules (via `PrepareProposal` and `ProcessProposal`) introduce implementation risks that could potentially lead to a chain halt.

cips/cip-tx-limits.md
rootulp: Agreed the motivation for the change is preventative.

> because every node can set the default to any value they choose to

Nodes can override the default but I expect most nodes will use the default. Assuming most nodes use the default config, transactions larger than 2 MiB won't be included in blocks so blob submitters will continue submitting blobs <= 2 MiB so that their blobs get included in blocks.

ninabarbakadze: I agree with you but it seemed like we wanted to have a hard limit on it.

rootulp: Yea, I'm wondering why did we want the hard limit? The non-consensus breaking change is still preventative.

cips/cip-28.md
jcstein, line 30:
```suggestion
Any changes to the block validity rules (via `PrepareProposal` and `ProcessProposal`) introduce implementation risks that could potentially lead to a chain halt.
```

cips/cip-28.md
```diff
@@ -0,0 +1,34 @@
...
+
+This CIP proposes to set the limit for transaction size. The proposal is to set the transaction size limit to 2MiB. Setting the transaction size limit is consensus-breaking.
+
+## Specification
+
+1. Transaction size is limited to 2MiB by setting the versioned parameter `MaxTxSize` to 2097152, which is 2MiB in bytes. From version v3 and above, in `CheckTx`, `PrepareProposal`, and `ProcessProposal`, each transaction's size is checked against the `appconsts.MaxTxSize` threshold. This ensures that transactions over the limit are rejected or excluded at all stages, from initial submission to execution.
+
+## Rationale
+
+The rationale for this proposal is to set the transaction size limit to 2MiB, even with 8MiB blocks, to prevent issues with gossiping large transactions. Gossiping an 8MiB transaction without chunking could be detrimental to the network. This is a consensus-breaking change.
```
jcstein, line 22:
```suggestion
The proposal to set the transaction size limit to 2MiB, even with blocks of 8MiB or larger, is primarily preventative. Gossiping transactions approaching 8MiB without chunking could be detrimental to the network. While this could be implemented as a non-consensus breaking change by modifying the default mempool `MaxTxBytes`, a consensus-breaking change was chosen for the following reasons:

1. Enforces a hard limit across all nodes, preventing discrepancies.
1. Provides stronger safeguards against potential network issues from large transactions.
1. Allows for controlled future experimentation with transaction sizes.

This approach prioritizes network-wide consistency and stability, but requires a coordinated network upgrade.
```

cips/cip-tx-limits.md
jcstein: Suggested revision below: https://github.com/celestiaorg/CIPs/pull/226#discussion_r1806974380

cips/cip-28.md
jcstein: Suggested revision below: https://github.com/celestiaorg/CIPs/pull/226#discussion_r1806968634

PR #228, opened 2024-10-19
docs: move CIPs 26,27,28 to Review

jcstein:


rootulp (approved):
CIP-27 and CIP-28 don't have test cases but it looks like that's optional per https://github.com/celestiaorg/CIPs/blob/0cf3a6fec91b145a5b96cdb5b217cfa7de358afd/cips/cip-1.md?plain=1#L287-L291 so LGTM.

PR #244, opened 2025-01-15
CIP 18: Add tx priority to requests 

cmwaters:
This PR makes a few minor modifications to CIP 18:  Standardised Gas and Pricing Estimation Interface

- Remove the reference implementation section. Client and server implementations in the light and consensus node respectively will have their own ADRs that eventually can be linked in this CIP
- Add an optional `TxPriority` field to indicate the urgency for a transaction to be finalized in a block. It has three levels: low, medium, and high.
- Modify the APIs to have two methods: requesting just the gas price and requesting the gas price and the gas. The latter will be used for non-PFB transactions.

### DIFF
--- a/cips/cip-18.md
+++ b/cips/cip-18.md
@@ -38,8 +38,8 @@ The following API is proposed for the standardised gas estimation service.
 
 ```proto
 service GasEstimator {
-    rpc EstimateFee(EstimateFeeRequest) returns (EstimateFeeResponse) {}
-    rpc EstimateFeeAndGas(EstimateFeeAndGasRequest) returns (EstimateFeeAndGasResponse) {}
+    rpc EstimateGasPrice(EstimateGasPriceRequest) returns (EstimateGasPriceResponse) {}
+    rpc EstimateGasPriceAndUsage(EstimateGasPriceAndUsageRequest) returns (EstimateGasPriceAndUsageResponse) {}
 }
 
 enum TxPriority {
@@ -49,20 +49,20 @@ enum TxPriority {
   HIGH = 3;
 }
 
-message EstimateFeeRequest {
+message EstimateGasPriceRequest {
     TxPriority tx_priority = 1;
 }
 
-message EstimateFeeResponse {
+message EstimateGasPriceResponse {
     double estimated_gas_price = 1;
 }
 
-message EstimateFeeAndGasRequest {
+message EstimateGasPriceAndUsageRequest {
     TxPriority tx_priority = 1;
     cosmos.tx.Tx tx = 2;
 }
 
-message EstimateFeeAndGasResponse {
+message EstimateGasPriceAndUsageResponse {
     double estimated_gas_price = 1;
     uint64 estimated_gas_used = 2;
 }
--- a/cips/cip-18.md
+++ b/cips/cip-18.md
@@ -43,8 +43,7 @@ service GasEstimator {
 }
 
 enum TxPriority {
-  // Default value must always be 0 in proto3
-  UNKNOWN = 0;
+  NONE = 0;
   LOW = 1;
   MEDIUM = 2;
   HIGH = 3;
--- a/cips/cip-18.md
+++ b/cips/cip-18.md
@@ -38,28 +38,50 @@ The following API is proposed for the standardised gas estimation service.
 
 ```proto
 service GasEstimator {
-    rpc EstimateGasUsage(EstimateGasUsageRequest) returns (EstimateGasUsageResponse) {}
-    rpc EstimateGasPrice(EstimateGasPriceRequest) returns (EstimateGasPriceResponse) {}
+    rpc EstimateFee(EstimateFeeRequest) returns (EstimateFeeResponse) {}
+    rpc EstimateFeeAndGas(EstimateFeeAndGasRequest) returns (EstimateFeeAndGasResponse) {}
 }
 
-message EstimateGasUsageRequest {
-    cosmos.tx.Tx tx = 1;
+enum TxPriority {
+  // Default value must always be 0 in proto3
+  UNKNOWN = 0;
+  LOW = 1;
+  MEDIUM = 2;
+  HIGH = 3;
 }
 
-message EstimateGasUsageResponse {
-    uint64 estimated_gas_used = 1;
+message EstimateFeeRequest {
+    TxPriority tx_priority = 1;
 }
 
-message EstimateGasPriceRequest {}
-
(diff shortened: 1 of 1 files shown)


cips/cip-18.md
```diff
@@ -38,49 +38,62 @@ The following API is proposed for the standardised gas estimation service.
...
-    rpc EstimateGasUsage(EstimateGasUsageRequest) returns (EstimateGasUsageResponse) {}
-    rpc EstimateGasPrice(EstimateGasPriceRequest) returns (EstimateGasPriceResponse) {}
+    rpc EstimateFee(EstimateFeeRequest) returns (EstimateFeeResponse) {}
+    rpc EstimateFeeAndGas(EstimateFeeAndGasRequest) returns (EstimateFeeAndGasResponse) {}
 }
 
-message EstimateGasUsageRequest {
-    cosmos.tx.Tx tx = 1;
+enum TxPriority {
+  NONE = 0;
```
rach-id, line 46: what would NONE represent? a default MEDIUM priority?

cips/cip-18.md
```diff
@@ -38,49 +38,62 @@ The following API is proposed for the standardised gas estimation service.
...
 }
 
-message EstimateGasPriceRequest {}
+message EstimateFeeResponse {
+    double estimated_gas_price = 1;
+}
+
+message EstimateFeeAndGasRequest {
+    TxPriority tx_priority = 1;
+    cosmos.tx.Tx tx = 2;
```
rach-id, line 62: would this just run simulateTx?

cips/cip-18.md
```diff
@@ -38,49 +38,62 @@ The following API is proposed for the standardised gas estimation service.
 
 ```proto
 service GasEstimator {
-    rpc EstimateGasUsage(EstimateGasUsageRequest) returns (EstimateGasUsageResponse) {}
-    rpc EstimateGasPrice(EstimateGasPriceRequest) returns (EstimateGasPriceResponse) {}
+    rpc EstimateFee(EstimateFeeRequest) returns (EstimateFeeResponse) {}
+    rpc EstimateFeeAndGas(EstimateFeeAndGasRequest) returns (EstimateFeeAndGasResponse) {}
```
rach-id, line 42: As per discussing with @vgonkivs, what do you think of adding a linear formula similar to PFBs that can be used by light clients as a backup if there is an issue with the gas estimation endpoint? It could be aggressive but at least ensures transaction inclusion most of the time.

cips/cip-18.md
```diff
@@ -38,49 +38,62 @@ The following API is proposed for the standardised gas estimation service.
...
 
 Given it's wide usage both in the Cosmos SDK and more broadly, the service would be implemented using gRPC. RESTful endpoints may optionally be supported in the future using something like `grpc-gateway`.
 
 Given the expected reliance on clients, all future changes must be done in a strictly non-breaking way.
 
+The notion of urgency, represented by the `TxPriority` field is both simple and subjective. Loosely it means:
+
+- High: whithin 1 block
+- Medium: within a few blocks
+- Low: withing the next hour
```
rach-id, line 81: Just to confirm what these mean from an implementation point of view (if this is the right place to discuss or we can leave this for an ADR):

- High: means returning the median gas of the highest 10% gas prices in the last 5 blocks, for example?
- Medium: would be the median of all gas prices in the last 5 blocks
- Low: would be the median of the lowest 10% gas prices in the last 5 blocks?

rach-id (commented):
implementation related questions that can be moved to an ADR if desired

cips/cip-18.md
vgonkivs: We could add it at least for the most popular txs.

cips/cip-18.md
```diff
@@ -38,49 +38,62 @@ The following API is proposed for the standardised gas estimation service.
...
-message EstimateGasUsageResponse {
-    uint64 estimated_gas_used = 1;
+message EstimateFeeRequest {
+    TxPriority tx_priority = 1;
 }
 
-message EstimateGasPriceRequest {}
+message EstimateFeeResponse {
+    double estimated_gas_price = 1;
+}
```
rootulp, line 58: [nit] I liked the previous names b/c `fee` implies gas price * gas and this response doesn't include an `estimated_gas` field

cips/cip-18.md
```diff
@@ -38,49 +38,62 @@ The following API is proposed for the standardised gas estimation service.
...
+message EstimateFeeAndGasRequest {
+    TxPriority tx_priority = 1;
+    cosmos.tx.Tx tx = 2;
+}
 
-message EstimateGasPriceResponse {
+message EstimateFeeAndGasResponse {
     double estimated_gas_price = 1;
+    uint64 estimated_gas_used = 2;
 }
```
rootulp, line 68: nit: fee and gas_price are used interchangeably here but [Cosmos docs](https://docs.cosmos.network/v0.46/basics/gas-fees.html#gas-and-fees) define them distinctly. Could this instead be named `EstimateGasPriceAndGas` or `EstimateFee` (IMO less preferable)

cips/cip-18.md
cmwaters: Yes the default which would be medium. Basically, I want tx priority to be optional (it doesn't have to be filled)

cips/cip-18.md
cmwaters: Yes

cips/cip-18.md
cmwaters: I wanted to include this in the ADR for celestia-node: PFBs would use the linear model, all other transactions would request the gas estimate

cmwaters: In the future we could add a model for other popular transactions like MsgSend, MsgTransfer, MsgClaim etc.

cips/cip-18.md
cmwaters: This should be covered in the celestia-app ADR

cips/cip-18.md
cmwaters: I had changed it to `fee` for brevity reasons but I got the same feeling around the accuracy of names as you. Maybe I will change it back and have it longer

cips/cip-18.md
rach-id: given that the simulation just uses the bytes, why not just provide the tx bytes instead of the `sdk.Tx` type?

cmwaters: You're right. I see it's been deprecated. Yeah let's change this to tx_bytes and the bytes format

PR #251, opened 2025-02-07
cip: prevent auto claiming of rewards. 

tac0turtle:
## Overview

This cip proposes a change to the distribution module, the change stops auto claiming of rewards when modifying delegation shares (delegation, redelgation, unbonding) and makes the movement of funds from the distribution module to a users account explicit.

### DIFF
--- a/cips/cip-30.md
+++ b/cips/cip-30.md
@@ -1,0 +1,108 @@
+| cip           | 30                                                                                         |
+|---------------|--------------------------------------------------------------------------------------------|
+| title         | Prevent Auto-Claiming of Staking Rewards                                                   |
+| description   | This CIP proposes removing the automatic claiming of staking rewards when modifying delegations. Instead, rewards are to be stored until manually claimed. |
+| author        | Dean Eigenmann ([@decanus](https://github.com/decanus)), Marko Baricevic ([@tac0turtle](https://github.com/tac0turtle)) |
+| discussions-to |                                 |
+| status        | Draft                                                                                      |
+| type          | Standards Track                                                                            |
+| category      | Core                                                                                       |
+| created       | 2025-02-07                                                                                 `|
+
+## Abstract
+
+Currently, when delegators modify their staking positions (e.g., redelegate or undelegate), the distribution module automatically claims all accrued staking rewards. This CIP proposes removing that automatic functionality, thereby allowing users to intentionally choose *when* to claim their rewards. Specifically, the distribution module will no longer auto-claim rewards on delegation-share changes. Instead, it will store the accrued rewards until a user explicitly calls `MsgWithdrawDelegatorReward`. This avoids unintended reward-claiming events that can trigger immediate tax or regulatory obligations.
+
+## Motivation
+
+Many delegators face tax or regulatory requirements forcing them to sell or otherwise handle any staking rewards the moment those rewards are claimed. Because the current system automatically claims rewards whenever a delegation is modified, users frequently have to deal with surprise or unintended taxable events, as well as unwanted administrative overhead. By decoupling reward claims from delegation operations, users can manage claiming events more intentionally.
+
+## Specification
+
+### Overview
+
+- **Existing Behavior**:
+  - When a delegator changes their delegation (e.g., via `MsgDelegate`, `MsgBeginRedelegate`, `MsgUndelegate`), the distribution module automatically triggers a reward claim for the delegator.
+  
+- **Proposed Behavior**:
+  - Remove or modify the distribution module’s “auto-claim” hook so that **no** rewards are automatically claimed on delegation changes.
+  - Delegators **only** claim their rewards by sending a `MsgWithdrawDelegatorReward` transaction (or the equivalent in the Celestia ecosystem).
+  - Pending rewards accumulate in the distribution module until explicitly claimed.
+
+### Detailed Design
+
+1. **Distribution Module Hook Changes**:
+   - Modify the existing hook(s) or condition(s) that trigger an automatic reward withdrawal during delegation modifications.
+   - Ensure that the reward calculation logic remains untouched (i.e., it continues to track and update the delegator’s reward pool share). Only the *withdrawal action* is changed.
+
+2. **Reward Storage**:
+   - The distribution module **continues** to store accrued rewards in the same internal accounting structures, just as it currently does before automatic withdrawals.
+   - No additional data structure is necessary beyond removing the forced withdrawal step.
+
+3. **Claiming Rewards**:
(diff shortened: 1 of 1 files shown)


cips/cip-30.md
```diff
@@ -0,0 +1,108 @@
...
+| discussions-to |                                 |
+| status        | Draft                                                                                      |
+| type          | Standards Track                                                                            |
+| category      | Core                                                                                       |
+| created       | 2025-01-27                                                                                 |
+
+## Abstract
+
+Currently, when delegators modify their staking positions (e.g., redelegate or undelegate), the distribution module automatically claims all accrued staking rewards. This CIP proposes removing that automatic functionality, thereby allowing users to intentionally choose *when* to claim their rewards. Specifically, the distribution module will no longer auto-claim rewards on delegation-share changes. Instead, it will store the accrued rewards until a user explicitly calls `MsgWithdrawDelegatorReward`. This avoids unintended reward-claiming events that can trigger immediate tax or regulatory obligations.
```
evan-forbes, line 14: seems like a simple enough change, do we have data on the rate of redelegations or undelegations or how much demand there is for this change? I couldn't think of any firm blockers to this, but its been a while since I dug through the maze that is the distribution and staking module logic.

separate to this limitted change,

>This is a simple change, if we wanted something more complex we could propose an LST like design

could we do, or has any sdk based chain done, claimless auto compounding, redelegation, undelegation without creating an LST? presumably this requires a significant change

cips/cip-30.md
```diff
@@ -0,0 +1,108 @@
+| cip           | 30                                                                                         |
+|---------------|--------------------------------------------------------------------------------------------|
+| title         | Prevent Auto-Claiming of Staking Rewards                                                   |
+| description   | This CIP proposes removing the automatic claiming of staking rewards when modifying delegations. Instead, rewards are to be stored until manually claimed. |
+| author        | Dean Eigenmann ([@decanus](https://github.com/decanus)), Marko Baricevic ([@tac0turtle](https://github.com/tac0turtle)) |
+| discussions-to |                                 |
+| status        | Draft                                                                                      |
+| type          | Standards Track                                                                            |
+| category      | Core                                                                                       |
+| created       | 2025-01-27                                                                                 |
```
evan-forbes, line 10: extremely optional nit, but this is a week old

cips/cip-30.md
evan-forbes, line 14: Would this require going back to (or keeping) a fork of the sdk using v0.50?

tac0turtle: yea i can get the data for the last few months. 

we have explored this, its a deeper change in the distribution module. I can look into this as an alternative.

cips/cip-30.md
tac0turtle: amended

cips/cip-30.md
tac0turtle: we may be able to get away without a fork, but due to the global registry we may need to maintain a small fork.

ValarDragon:
Even independent of taxation, this is just generally a massive UX win. Its very annoying to have to explain or track as is. (Note, spec of this was definitely in large part my fault for not thinking of the UX pain here)

cips/cip-30.md
```diff
@@ -0,0 +1,108 @@
+| cip           | 30                                                                                         |
+|---------------|--------------------------------------------------------------------------------------------|
+| title         | Prevent Auto-Claiming of Staking Rewards                                                   |
+| description   | This CIP proposes removing the automatic claiming of staking rewards when modifying delegations. Instead, rewards are to be stored until manually claimed. |
+| author        | Dean Eigenmann ([@decanus](https://github.com/decanus)), Marko Baricevic ([@tac0turtle](https://github.com/tac0turtle)) |
+| discussions-to |                                 |
```
musalbas, line 6: Can we add discussion link?

tac0turtle: added

cips/cip-30.md
```diff
@@ -0,0 +1,108 @@
...
+## Specification
+
+### Overview
+
+- **Existing Behavior**:
+  - When a delegator changes their delegation (e.g., via `MsgDelegate`, `MsgBeginRedelegate`, `MsgUndelegate`), the distribution module automatically triggers a reward claim for the delegator.
+  
+- **Proposed Behavior**:
+  - Remove or modify the distribution module’s “auto-claim” hook so that **no** rewards are automatically claimed on delegation changes.
+  - Delegators **only** claim their rewards by sending a `MsgWithdrawDelegatorReward` transaction (or the equivalent in the Celestia ecosystem).
```
rootulp, line 29: Celestia hasn't overriden the default x/distribution module so this seems unnecessary.

```suggestion
  - Delegators **only** claim their rewards by sending a `MsgWithdrawDelegatorReward` transaction.
```

cips/cip-30.md
```diff
@@ -0,0 +1,108 @@
...
+-
++    // New behavior: do nothing related to claiming.
++    // Rewards remain in the distribution module until user calls WithdrawDelegatorReward explicitly.
+ }
+```
+
+## Security Considerations
+
+ • Users holding large amounts of unclaimed rewards could become more attractive targets for potential exploits if a vulnerability existed in the distribution module. However, this scenario already exists between automatic claim intervals or across multiple validators.
+ • Removing auto-claim does not introduce new vectors for abuse; it merely defers the claim to a user-initiated transaction.
```
rootulp, line 104: [nit]

```suggestion
- Users holding large amounts of unclaimed rewards could become more attractive targets for potential exploits if a vulnerability existed in the distribution module. However, this scenario already exists between automatic claim intervals or across multiple validators.
- Removing auto-claim does not introduce new vectors for abuse; it merely defers the claim to a user-initiated transaction.
```

rootulp (approved):
LGTM, thanks!

cips/cip-30.md
```diff
@@ -0,0 +1,108 @@
+| cip           | 30                                                                                         |
```
jcstein, line 1:
```suggestion
| cip           | 29                                                                                         |
```

jcstein:
it looks like editors don't have permission to edit this branch @tac0turtle, can you please grant permission? thank you

cips/cip-30.md
tac0turtle: so changing to be native compounding can be done, its not that hard. The main change is the UX around, users will need to undelegate to claim rewards. This requires a conversation on UX before moving forward. The nice part is that we would reduce state machine complexity per block significantly.

cips/cip-30.md
```diff
@@ -0,0 +1,108 @@
+| cip           | 29                                                                                         |
```
jcstein, line 1:
```suggestion
| cip           | 30                                                                                         |
```

based on our discussion @tac0turtle 🫡

jcstein (approved):

Issue #120, opened 2024-04-03
CIP-12 Security Considerations


rootulp:
## Context

https://github.com/celestiaorg/CIPs/blob/46e14992353f63fb9c0a7553f33881947688ec17/cips/cip-12.md?plain=1#L583-L587

## Problem

This section of CIP-12 still contains the template boilerplate.

## Proposal

Replace the boilerplate with content specific to CIP-12.

rootulp:
I wrote a [forum post](https://forum.celestia.org/t/cip-relayer-incentivisation-middleware/1383/6) which took a first stab at fleshing out some security considerations so feel free to use or not use any of that content in the CIP.

cc: @womensrights

rootulp:
Can close this as won't do if we merge https://github.com/celestiaorg/CIPs/pull/257

cc: @womensrights

PR #257, opened 2025-02-24
docs(cip-12): withdraw

rootulp:
Closes https://github.com/celestiaorg/CIPs/issues/120 because we don't need to add security considerations if we withdraw the proposal.

## Description
Withdraw CIP-12 because ICS-29 will be removed from ibc-go v10. See https://celestia-team.slack.com/archives/C047X1T6ERK/p1740423665065499

cc: @womensrights @AdityaSripal


rootulp:
Markdownlint is failing for an unrelated file.

jcstein (approved):

rootulp:
Let's hold off on merging until one of the authors approves.

Issue #204, opened 2024-09-11
cip-13: add v2 params


rootulp:
## Context

celestia-app app version 2 is expected to be activated on Mainnet on September 18. 

## Problem

https://github.com/celestiaorg/CIPs/blob/main/cips/cip-13.md is a living document so I think the intention was that it should be updated to reflect the values of params on mainnet.

Since app version 2 introduces new params (ICA, PFM, minfee), cip-13 will need to be udpated.

### Option A 
Deprecate CIP-13 in favor of referencing https://celestiaorg.github.io/celestia-app/parameters.html

### Option B 
Update CIP-13 to add new params introduced by app version 2.
- [ ] https://github.com/celestiaorg/CIPs/issues/113

jcstein:
In favor of Option A, because it's already always documented at https://celestiaorg.github.io/celestia-app/parameters.html

PR #115, opened 2024-03-27
docs(cip-13): add NetworkMinGasPrice to params list

ninabarbakadze:
## Overview

- Fixes #113

Issue #113, opened 2024-03-21
Update CIP-13 to add minfee.GlobalMinGasPrice


ninabarbakadze:
## Description

New gov-modifiable parameter `minfee.GlobalMinGasPrice` has been introduced recently [in this pr](https://github.com/celestiaorg/celestia-app/pull/3094). Add it to the CIP's module parameters list.

PR #259, opened 2025-03-04
docs: withdraw cip-13

rootulp:
CIP-13 is a maintenance burden and hasn't been updated to include new parameters added in v2 or v3. It also hasn't been updated to account for governance modifiable changes. This PR withdraws the CIP which seems safe because it never progressed past **Draft**.

Closes https://github.com/celestiaorg/CIPs/issues/204 
Closes https://github.com/celestiaorg/CIPs/issues/113
Closes https://github.com/celestiaorg/CIPs/pull/115


jcstein (commented):
SGTM, agreed it is an unecessary burden as this is already tracked at celestiaorg.github.io/celestia-app/parameters_v3.html

jcstein (approved):

PR #271, opened 2025-03-17
docs(cip-33): add prerequisite CIPs

rootulp:
Oops I didn't notice that the `requires` field in the table was empty


jcstein (approved):
same. LGTM

PR #272, opened 2025-03-17
refactor(README): remove extra author identifiers

rootulp:
Extremely optional PR to make the README markdown tables easier to maintain.
1. Remove names and emails because every CIP author must have a Github username. IMO it's sufficient to identify authors by one thing (Github username) instead of multiple things (Github username, name, email).
2. Make the tables left aligned
3. Reduce the width of columns so it's easier to edit the table


rootulp:
> my only nit would be a question of whether or not we're going to change this on CIPs?

Yea can do. We can also update the template to include just Github username.

PR #274, opened 2025-03-17
docs(cip-31): move to review

rootulp:
Is there a reference implementation I can add to the detailed design / reference implementation section? @tac0turtle


tac0turtle:
i have a link to the code but its quite large to include here, what is the preferred path?

jcstein:
this is the current guidance: https://cips.celestia.org/cip-001.html#linking-to-external-resources

happy to change this too if we think it's safe

jcstein (approved):

PR #279, opened 2025-03-21
cip: Header Pruning for LNs

Wondertan:
## Overview

CIP for Header Pruning. As Header Pruning breaks users, particularly by preventing historical queries(to be fixed in subsequent CIP), it was decided to promote this change to a CIP. 

This CIP start a series of DA targeted CIPs to optimize bandwidth and storage usage for LN, particularly targeting overhead brought by headers.

### DIFF
--- a/cips/README.md
+++ b/cips/README.md
@@ -74,6 +74,7 @@ Read [CIP-1](./cip-001.md) for information on the CIP process.
 | [32](./cip-032.md) | Add Hyperlane to Celestia                                                    | [@c-node](https://github.com/S1nus), [@cmwaters](https://github.com/cmwaters)                                                                                                |
 | [33](./cip-033.md) | Lotus Network Upgrade                                                        | [@evan-forbes](https://github.com/evan-forbes)                                                                                                                               |
 | [34](./cip-034.md) | Lower Sampling and Pruning Window to 14 days                                | [@cmwaters](https://github.com/cmwaters), [@Nashqueue](https://github.com/Nashqueue)                                                                                         |
+| [35](./cip-035.md) | Header Pruning for Light Nodes                                             | [@Wondertan](https://github.com/Wondertan)                                                                                                                                   |
 
 ## Contributing
 
--- a/cips/SUMMARY.md
+++ b/cips/SUMMARY.md
@@ -37,6 +37,7 @@
   - [CIP-32](./cip-032.md)
   - [CIP-33](./cip-033.md)
   - [CIP-34](./cip-034.md)
+  - [CIP-35](./cip-035.md)
 
 - [Core Devs Call notes](./notes/README.md)
   - [CDC #14](./notes/cdc-14.md)
--- a/cips/cip-035.md
+++ b/cips/cip-035.md
@@ -1,4 +1,4 @@
-| cip | XX (assigned by Editors) |
+| cip | 35 |
 | - | - |
 | title | Header Pruning for Light Nodes |
 | description | Mechanism to retain a fixed range of headers instead of whole history |


cips/cip-x.md
```diff
@@ -0,0 +1,76 @@
+| cip | XX (assigned by Editors) |
+| - |  |
+| title | Header Pruning for LNs |
```
jcstein, line 3:
```suggestion
| title | Header Pruning for Light Nodes |
```

jcstein: IMO best to define acronym before using it

cips/cip-x.md
```diff
@@ -0,0 +1,76 @@
...
+| author | Hlib Kanunnikov ([@Wondertan](https://github.com/Wondertan)) |
+| discussions-to | <https://forum.celestia.org/t/cip-header-pruning-for-lns/1958> |
+| status | Draft |
+| type | Standards Track |
+| category | Data Availability |
+| created | 2025-03-21 |
+
+## Abstract
+
+Currently, every DA node type synchronizes all historical headers starting from genesis (or other statically configured
```
jcstein, line 14:
```suggestion
Currently, every data availability (DA) node type synchronizes all historical headers starting from genesis (or other statically configured
```

strictly for clarity

cips/cip-x.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+## Parameters
+
+| Parameter           | Proposed value | Description                                                            | Changeable via Governance |
+|---------------------|----------------|------------------------------------------------------------------------|---------------------------|
+| HeaderPruningWindow | 14 days        | Defines how many headers to retain head backwards. Configurable per LN | No                        |
+
+By default, HeaderPruningWindow equals to Availability/SamplingWindow and can be modified by LN operators.
+// TODO: Point to the right parameter with CIP?
```
jcstein, line 54: is this R4R?

cips/cip-x.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Rationale
+
+Initially, we add pruning support only to LNs to ensure reliable header retrieval for historical headers. Otherwise,
+nodes configured with larger [HeaderPruningWindow](#parameters) than in connected FN/BNs won’t be able to retrieve
```
jcstein, line 59:
```suggestion
nodes configured with larger [HeaderPruningWindow](#parameters) than in connected full nodes (FN)/bridge nodes (BNs) won’t be able to retrieve
```

cips/cip-x.md
Wondertan: It is. This a question for reviewers. Whether its necessary link this somewhere.

Sidenote: there is still a bit of mess with those params. Some of them needs to be changed anr some needs to be defined in a cip

jcstein: where do the params live now?

Wondertan: In code :joy:

cips/cip-x.md
```diff
@@ -0,0 +1,76 @@
...
+	estimatedCutoffHeight = min(tail.Height() + estimatedRange, head.Height())
+	return estimatedCutoffHeight
+}
+```
+
+## Parameters
+
+| Parameter           | Proposed value | Description                                                            | Changeable via Governance |
+|---------------------|----------------|------------------------------------------------------------------------|---------------------------|
+| HeaderPruningWindow | 14 days        | Defines how many headers to retain head backwards. Configurable per LN | No                        |
```
rootulp, line 51: [nit] AFAICT the HeaderPruningWindow is a duration if time not a number of headers.

```suggestion
| HeaderPruningWindow | 14 days        | Defines the length of time to retain headers. Configurable per LN | No                        |
```

cips/cip-x.md
```diff
@@ -0,0 +1,76 @@
...
+## Rationale
+
+Initially, we add pruning support only to LNs to ensure reliable header retrieval for historical headers. Otherwise,
+nodes configured with larger [HeaderPruningWindow](#parameters) than in connected FN/BNs won’t be able to retrieve
+needed headers. We can revisit this consideration if header storage overhead ever becomes pressing for FN/BNs.
+
+## Backwards Compatibility
+
+The proposed change is backwards compatible with the current implementation on the protocol. However, LN operators
+enabling pruning will loose the ability to retrieve historical headers beyond the [HeaderPruningWindow](#parameters).
```
rootulp, line 65:
```suggestion
enabling pruning will lose the ability to retrieve historical headers beyond the [HeaderPruningWindow](#parameters).
```

rootulp (commented):
LGTM after resolving the parameter question. Given the DA network parameters aren't already in CIPs (or specs) somewhere, I suggest keeping the source of truth in code and removing the parameter from this CIP.

The rationale is that it's way easier to update the parameter in code and there is no risk of the parameter in docs becoming stale with the implementation in code.

cips/cip-x.md
musalbas: The current pruning window in CIP-4 is 30 days not 14 days, imo it should be kept the same until if/when there's a CIP to reduce it.

Wondertan:
>  Given the DA network parameters aren't already in CIPs (or specs) somewhere, I suggest keeping the source of truth in code and removing the parameter from this CIP.

@rootulp, rhere is SamplingWindow param that is already in CIP-4. We can tie HeaderPruningWindow to it which is 30 days for simplicity. It should though be defined based of TrustingPeriod which is 14 days(basing on @nashqueue's findings), but that's gonna be a different CIP by someone else(like @cmwaters) that will define the TrustingPeriod and resolve issues in between all the params by adjusting them.

So given we have the linkable parameter in the old CIP, I believe it fine to define the new parameter in the CIP.

cmwaters:
Yeah there are already node parameters defined in CIP 4 (and we may look to adjust those soon). These are important parameters so I would also prefer to have them documented here (possibly in the same way we do core/app parameters)

cips/cip-x.md
```diff
@@ -50,7 +50,8 @@ func estimateTail(head, tail Header, blockTime, pruningWindow time.Duration) (he
 |---------------------|----------------|-----------------------------------------------|---------------------------|
 | HeaderPruningWindow | 30 days        | Defines the length of time to retain headers  | No                        |
 
-By default, HeaderPruningWindow equals to SamplingWindow from [CIP-004](./cip-004.md), but its configurable per LN.
+By default, HeaderPruningWindow equals to SamplingWindow from [CIP-004](./cip-004.md). It can be modified per LN, but
+it must be more or equal to the SamplingWindow.
```
renaynay, line 54: Do you want to clarify why we don't want to allow LN to have pruning window < sampling window?

Wondertan: > As there is currently way to sync history beyond HPW, the HPW less than SPW is gonna fail sampling. SPW is currently 30 days.

From notion. Doesn't hurt to add this here

Wondertan: [ceed598](https://github.com/celestiaorg/CIPs/pull/279/commits/ceed5986e1e2cf9bedad74dda89455957bead4ed)

jcstein (approved):
LGTM

cips/cip-x.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Motivation
+
+Light nodes (LNs) store the whole history of ExtendedHeaders in their entirety, including ValidatorSet, Commit and data availability header (DAH). The motivation
+is to decrease LN storage requirements by retaining only a configurable window of recent headers.
+
+## Specification
+
+In order to reference the last stored header, we introduce a notion of **Tail** header of the chain, complementing
+**Head** header. Tail header height is estimated with [HeaderPruningWindow](#parameters), block time and Head height.
```
ebuchman, line 26: "estimated" because pruning window is defined by time rather than height?

cips/cip-x.md
```diff
@@ -0,0 +1,77 @@
...
+
+- On the first start of a node, the Tail is retrieved from trusted peers alongside with the Head, performing subjective
+initialization.
+  - Once retrieved, Tail becomes the point of trust for the initializing node where everything beforehand assumed valid
+ (unless syncing from genesis)
+  - Head, on the other hand, is not trusted and becomes a synchronization target for the node.
```
ebuchman, line 32: if Head was retrieved in the exact same manner as Tail as part of subjective init, and Tail is trusted, what's the justification for not trusting Head? I would have expected the latest height that is subjectively initialized to serve as the first Tail

cips/cip-x.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Abstract
+
+Currently, every data availability (DA) node type synchronizes all historical headers starting from genesis (or other statically configured
+historical header) until the subjectively initialized head of the chain. We change that by adding a way to sync a
+constant size range of headers instead of the whole history.
+
+## Motivation
+
+Light nodes (LNs) store the whole history of ExtendedHeaders in their entirety, including ValidatorSet, Commit and data availability header (DAH). The motivation
```
ebuchman, line 20: What is the motivation for LNs to keep any historical data in the first place?

cips/cip-x.md
```diff
@@ -0,0 +1,77 @@
...
+is to decrease LN storage requirements by retaining only a configurable window of recent headers.
```
ebuchman, line 21: How should one reason about what an appropriate configuration for this should be?

cips/cip-x.md
```diff
@@ -0,0 +1,77 @@
...
+ (unless syncing from genesis)
+  - Head, on the other hand, is not trusted and becomes a synchronization target for the node.
+  - All the headers from Tail to Head are then retrieved and verified using signature-based forward verification.
+- As the chain progresses, nodes continuously re-estimate Tail and cut off stored headers beyond the new Tail, retaining
+roughly [HeaderPruningWindow](#parameters) amount of headers.
+
+The estimation of Tail height is done as follows:
+
+```go
+func estimateTail(head, tail Header, blockTime, pruningWindow time.Duration) (height uint64) {
```
ebuchman, line 40: this is confusing, why is "now" involved? I would have expected it to be:

```
numBlockToKeep = pruningWindow / blockTime
newTail = head - numBlocksToKeep
return max(currentTail, newTail) // in case newTail is somehow older than currentTail
```

cips/cip-x.md
ebuchman: i see this is basically alluded to in the parameter section but it would be nice if it was clear up front in the motivation

cips/cip-x.md
```diff
@@ -0,0 +1,77 @@
...
+	estimatedCutoffHeight = min(tail.Height() + estimatedRange, head.Height())
+	return estimatedCutoffHeight
+}
+```
+
+## Parameters
+
+| Parameter           | Proposed value | Description                                   | Changeable via Governance |
+|---------------------|----------------|-----------------------------------------------|---------------------------|
+| HeaderPruningWindow | 30 days        | Defines the length of time to retain headers  | No                        |
```
ebuchman, line 51: this isnt really enforcable anywhere right? do we have a  way to distinguish between parameters that are enforceable as part of consensus and those that are just local configurations but are nonetheless important for security?

cips/cip-x.md
```diff
@@ -0,0 +1,77 @@
...
+
+The proposed change is backwards compatible with the current implementation on the protocol. However, LN operators
+enabling pruning will lose the ability to retrieve historical headers beyond the [HeaderPruningWindow](#parameters).
+Therefore, we will keep pruning opt-in and keep the ability to initialize from genesis or any arbitrary header in the
+past. This way node operators can benefit from the storage savings if they know from which point they can safely prune.
+Future protocol upgrade will solve this limitation by introducing an efficient way to retrieve historical headers.
+
+## Security Considerations
+
+This changes does not affect security of the protocol and its participants and is strictly an optimization for LN storage.
```
ebuchman, line 73:
```suggestion
This changes does not affect security of the protocol and its participants and is strictly an optimization for LN storage, assuming LN's respect the constraint `HeaderPruningWindow >= SamplingWindow`.
```

cips/cip-x.md
```diff
@@ -0,0 +1,77 @@
...
+}
+```
+
+## Parameters
+
+| Parameter           | Proposed value | Description                                   | Changeable via Governance |
+|---------------------|----------------|-----------------------------------------------|---------------------------|
+| HeaderPruningWindow | 30 days        | Defines the length of time to retain headers  | No                        |
+
+By default, HeaderPruningWindow (HPW) equals to SamplingWindow (SW) from [CIP-004](./cip-004.md). It can be modified per LN, but
```
ebuchman, line 53: minor nit but unfortunately the literal `SamplingWindow` doesnt show up in CIP-004 even though it's defined implicitly

ebuchman (commented):
Nice. Some nits and a few stupid questions

cips/cip-x.md
Wondertan: The intention is to cross-check that Tail matches the Head. We do have this point of trust and we could also trust both of these headers, but then we introduce an additional assumption that we can prevent.

Requesting the Head is actually optional, as we gonna get the most recent one with gossiping, but it may trigger syncing a bit earlier.

I should rephrase this part as I got the same question in sync.

cips/cip-x.md
Wondertan: There is/was none. It's a techdebt that became a product issue and the time has come to clean it up.

cips/cip-x.md
Wondertan: Do you mean distinguishable in CIP process? I don't think we have it, but might be worth introducing.
cc @jcstein

cips/cip-x.md
Wondertan: Good one!

cips/cip-x.md
Wondertan: What do you mean by "doesn't show up?"

Wondertan:
@jcstein, when should we assign the number

cips/cip-x.md
Wondertan: This snippet is actually a wrong one. This one is for for finding the range that needs to trimmed off. Thanks for catching that!

The now is used there as the starting point as opposed to the time of the last header. This is inherited from pruning for samples as its how it works currently. However, thinking about this now in the case of header pruning, I think it's reasonable to just cut off basing of the most recent header.

Wondertan:
GH applied the same suggestion from Ethan 5 times as separate commits lol

cips/cip-x.md
ebuchman: Right. Not a huge deal but I'd include that in the motivation along with the need to store at least the SampleWindow

cips/cip-x.md
ebuchman: Yeh that's all I meant. Looks like parameters have moved out of CIP into https://celestiaorg.github.io/celestia-app/parameters.html so maybe there should just be a section there for the non consensus params

cips/cip-x.md
ebuchman: i mean ctrl-f for `SamplingWindow` shows nothing because its all "Sampling Window". a bit pedantic but just from a searchability perspective 😅 

also not seeing SamplingWindow in the https://celestiaorg.github.io/celestia-app/parameters.html

cips/cip-x.md
jcstein: > this isnt really enforcable anywhere right?

is it enforceable on celestia-node within the code @Wondertan ? are parameters like this best to leave documented in the code or should we have them publicly accessible elsewhere similar to the celestia-app specs?

> Looks like parameters have moved out of CIP into https://celestiaorg.github.io/celestia-app/parameters.html 

yeah, these have moved and CIP-13 is withdrawn.

cips/cip-x.md
jcstein: > also not seeing SamplingWindow in the https://celestiaorg.github.io/celestia-app/parameters.html

similarly to comment above, this is because the parameters linked are specific to celestia-app. celestia-node parameters still live in code and not yet in the specs. cc @Wondertan and @walldiss here to get this implemented https://github.com/celestiaorg/celestia-node/tree/main/specs/src

jcstein:
@Wondertan after comments are resolved. can you please close the ones above that you've resolved with a link to the commit before resolving? thank you!

then we'll assign number and get this merged. just want to make sure @ebuchman's comments are addressed!

cips/cip-x.md
Wondertan: These types of parameter should be publicly specified as there are security and product implications, but there are not consensus/state machine enforcable.

cips/cip-x.md
Wondertan: 7bb5933b4abd314c272a4f711a78999a41106d11

ebuchman: fixed a typo but otherwise lgtm

cips/cip-x.md
jcstein: understood. let's link directly to CIP-4 for that from here to link together `HeaderPruningWindow` and `Sampling Window` and if we use `SamplingWindow` we should edit CIP-4

jcstein:
this should be good after comments are resolved @Wondertan

cips/cip-x.md
Wondertan: Something I realized coming from the actual implementation:
* it should take `trustingPeriod` instead of `headerPruningWindow` in here. As HPR can be arbitrary increased, it can become bigger then TP and more importantly the unbonding window, which then becomes a security issue due to LRA.
* Using `head` in tail estimation actually make `tail` requesting dependent on `head`. This is not ok, because it adds a round trip of delay for the subjective init. So using `now` should solve this while trading off a bit of accuracy in estimation. But that should be fine as long TP is noticeably less then UW

cips/cip-x.md
Wondertan: > here to link together HeaderPruningWindow and Sampling Window

They are linked in Motivation and Parameters sections.

> if we use SamplingWindow we should edit CIP-4

Not sure why?

jcstein: > Not sure why?

Not necessary to do here now, I was proposing it as a way to resolve @ebuchman 's first comment

jcstein:
I just realized maybe @Wondertan doesn't have permissions to resolve the comments. can you confirm?

cips/cip-x.md
```diff
@@ -0,0 +1,79 @@
...
```
jcstein, line 55:
```suggestion
By default, HeaderPruningWindow (HPW) is equal to SamplingWindow (SW) from [CIP-004](./cip-004.md). It can be modified per LN, but
```

cips/cip-x.md
Wondertan: Something I realized coming from the actual implementation:
* It should take `trustingPeriod` instead of `headerPruningWindow` in here. As HPR can be arbitrary increased, it can become bigger then TP and more importantly the unbonding window, which then becomes a security issue due to LRA.  See [fa9243f](https://github.com/celestiaorg/CIPs/pull/279/commits/fa9243f39d534d33721c42ceaeba33c25be91aea)
* Using `head` in tail estimation actually make `tail` requesting dependent on `head`. This is not ideal, because it adds a round trip of delay for the subjective init. You first request Head and then the Tail. This is kinda unavoidable, because we need some header to be time reference to convert time to heights. We could avoid this by hardcoding the time of genesis header or some other header, but that might be a bit too much for now.

cips/cip-x.md
```diff
@@ -0,0 +1,85 @@
+| cip | XX (assigned by Editors) |
```
jcstein, line 1:
```suggestion
| cip | 35 |
```

cips/cip-x.md
ebuchman: 1. not sure I follow here. Isn't SamplingWindow (~1month) already bigger than TrustingPeriod (~two weeks)? And we need pruningWindow >= SamplingWindow.  I would think the light client syncing should be independent of what's happening here. If the LC hasn't been online in at least TP, it will need to resubjectively initialize, but once it does it can safely go backwards via hash links to fill in older blocks

2. per above, my sense is a light node should first sync, and then decide if it needs to backfill. how is subjective init happening anyways?

Wondertan: > Isn't SamplingWindow (~1month) already bigger than TrustingPeriod (~two weeks)?

Yes, but that's exactly where the problem is. The fact that SamplingWindow and subsequently bigger then TrustingPeriod imposes risks on newly initialized nodes. If we start syncing from the Tail beyond TP we are in trouble due to LR and the change here is to prevent that.

Wondertan: > how is subjective init happening anyways?

So the missing context here might be that we do subj init a bit differently compared to other PoS projects. We automate that in a way that I believe it makes the process even less error prone. Usually, users are asked to enter on subj init a trusted hash of a header with validator set they believed to be correct. Instead, we introduce notion of a trusted peer that we request such a header with valset from. This in turn removes the hassle from user to find the correct hash, improving UX and reduces risks of phishing and other rookie mistakes by users pasting wrong hashes. The trusted peers are by default hardcoded infrastructure nodes, that can be altered in configs.

So my point there is that now instead of a single RT, there is gonna be two for two headers. This is okayish, but not ideal

ebuchman: i see, in that case shoudn't we just get a very recent header from the trusted peer for subjective init, possibly even the latest head, and then calculate what tail we need from that, and sync backwards? syncing backwards doesnt suffer from LR attacks, only syncing forward from something outside the TP

cips/cip-x.md
```diff
@@ -0,0 +1,85 @@
...
+
+The estimation of Tail is done as follows:
+
+```go
+func estimateTail(head Header, blockTime, window time.Duration) (height uint64) {
+    headersToRetain := window / blockTime
+    tail := head.Height() - headersToRetain
+    return tail
+}
+```
```
walldiss, line 47: Re-posting thought I've shared with you on 1-1:
The current estimation approach relies on a constant blockTime parameter, but in reality, block times can vary. It means that estimated Tail could drift from the intended SamplingWindow timeframe.
Given that `HeaderPruningWindow` must be ≥ `SamplingWindow` for proper sampling it can lead to potential sampling failures if too few headers are retained.
If the estimation is too conservative (retains more headers than needed), it reduces the storage efficiency benefits

Alternative approach:
Instead of local estimation, consider having nodes request the exact Tail header from trusted peers. The trusted peer could find the header closest to time.now() - SamplingWindow, providing precise alignment with sampling requirements and eliminating estimation errors.

This approach seems more reliable while still achieving the storage optimization goals. Thoughts?

cips/cip-x.md
Wondertan: Backwards sync is the next stage and something that was on our radar for quite a while. It gonna help us to avoid syncing validator sets and commits which are necessary for forward sync. We also gonna do it with MMR so we can proof inclusion in the past without needing the whole header chain up to that point in the past.

cips/cip-x.md
```diff
@@ -0,0 +1,85 @@
...
+| status | Draft |
+| type | Standards Track |
+| category | Data Availability |
+| created | 2025-03-21 |
+
+## Abstract
+
+Currently, every data availability (DA) node type synchronizes all historical headers starting from genesis (or other statically configured
+historical header) until the subjectively initialized head of the chain. We change that by adding a way to sync a
+constant size range of headers instead of the whole history.
```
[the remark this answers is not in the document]
Wondertan: It more length/size instead of time. Constant time usually refers to O(1) algorithms, which is not the case here.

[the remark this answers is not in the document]
walldiss: Thanks for clarifying. I see where the confusion came from—the term "constant time" I used wasn’t referring to algorithmic complexity (O(1)), but rather to the fixed duration (e.g., 30 days) used to define the pruning window, as mentioned in the CIP itself.

To avoid ambiguity, perhaps it might be clearer to explicitly state "a constant duration (time-based window)" or "fixed-duration window" rather than "constant size," since the actual number of headers stored will vary slightly depending on the block production rate.

Would wording like "fixed-duration range of headers" better reflect your intended meaning?

cips/cip-x.md
Wondertan: After the multiple sync and async discussions on this matter, we decided to leave the estimation logic as it is for the subjective initialization. However, we uncovered a bunch of conflict issues between header pruning and data/sampling pruning on the implementation side as an output of this point, so thank you for raising it!

cips/cip-x.md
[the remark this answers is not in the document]
Wondertan: Ok, I can clarify it there

Wondertan:
This is RFR again

jcstein (approved):

Issue #284, opened 2025-04-02
cip-31: describe which account types are affected


rootulp:
CIP-31 needs to be updated to describe the vesting account types that it:
- applies to: continuous and delayed
- does not apply to: permanent and periodic

cc: @tac0turtle

rootulp:
https://github.com/celestiaorg/cosmos-sdk/pull/438 now removes the ability to create permanent and periodic vesting accounts so that should be reflected in the CIP.

PR #290, opened 2025-04-07
CIP-31: updates according to the implemention

tac0turtle:
## Overview

This pr updates the cip-31 guidelines and implementation in order to make the design and flow as simple as possible. 

closes https://github.com/celestiaorg/CIPs/issues/284

### DIFF
--- a/cips/cip-031.md
+++ b/cips/cip-031.md
@@ -32,42 +32,45 @@ By integrating staking rewards directly into the lockup schedule, this proposal 
 
 ### Core Mechanism
 
-1. **Account Type Verification:**
+#### Account Type Verification
+
    When staking rewards are distributed (e.g., via `MsgWithdrawDelegatorReward`), the distribution module will first check if the receiving account is a lockup account.
 
-   - **If it is:**
-     - Verify that it is of a modifiable lockup type.
-     - Calculate the new lockup amount:
+- **If it is:**
+  - Verify that it is of a modifiable lockup type.
+  - Calculate the new lockup amount:
        `New Lockup Amount = Original Lockup Amount + Claimed Rewards`
-     - Determine the new unlock rate by dividing the updated lockup amount by the remaining lockup period.
-     - Update the lockup schedule and the account balance accordingly.
-   - **If it is not a lockup account:**
-     - Process the reward as a normal reward transfer.
+  - Determine the new unlock rate by dividing the updated lockup amount by the remaining lockup period.
+  - Update the lockup schedule and the account balance accordingly.
+- **If it is not a lockup account:**
+  - Process the reward as a normal reward transfer.
 
-2. **Reward Claim Paths:**
+#### Reward Claim Paths
+
    The CIP covers multiple paths where rewards might be integrated:
 
-   - **Direct Reward Claim:**
+- **Direct Reward Claim:**
      The user invokes a reward withdrawal message and the lockup schedule is updated.
 
-   - **Redelegation and Unbonding:**
+- **Redelegation and Unbonding:**
      Hooks in the staking module ensure rewards are claimed and added to the lockup schedule prior to processing redelegation or unbonding.
-     - *Note: IF [CIP-30](https://github.com/celestiaorg/CIPs/pull/251) is accepted this path is no longer available.*
+  - *Note: IF [CIP-30](https://github.com/celestiaorg/CIPs/pull/251) is accepted this path is no longer available.*
 
-   - **Validator Commission:**
+- **Validator Commission:**
      Commission rewards are split from staking rewards: commissions go directly to the account (unlocked) while staking rewards follow the lockup schedule.
 
-3. **Token State Management:**
+#### Token State Management
+
    Tokens are tracked in four states:
 
-   - Locked & Unstaked
-   - Locked & Staked
-   - Unlocked & Unstaked
-   - Unlocked & Staked
+- Locked & Unstaked
+- Locked & Staked
+- Unlocked & Unstaked
+- Unlocked & Staked
 
-4. **Example Implementations:**
+#### Example Implementations
 
-### Continuous Vesting Accounts
+##### Continuous Vesting Accounts
 
    Consider a lockup schedule initially granting 100 tokens over 365 days (unlocking ≈0.274 tokens/day). Assume 30% of the lockup period has passed (109.5 days), meaning 30 tokens have already unlocked, and 70 tokens remain locked.
 
@@ -83,7 +86,7 @@ By integrating staking rewards directly into the lockup schedule, this proposal 
 
    A portion of the reward corresponding to the elapsed vesting duration (30% in this case) unlocks immediately, while the rest (70%) is added to the future vesting schedule.
 
-### Delayed Vesting Accounts
+##### Delayed Vesting Accounts
 
    For a `DelayedVestingAccount`, the entire principal vests at a single point in the future (`EndTime`).
    Consider an account set up with 1000 tokens scheduled to unlock entirely after 730 days (2 years).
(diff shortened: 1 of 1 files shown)


cips/cip-031.md
```diff
@@ -67,48 +66,83 @@ By integrating staking rewards directly into the lockup schedule, this proposal
    - Unlocked & Staked
 
 4. **Mathematical Model & Example:**
```
rootulp, line 68: This list item is no longer needed

```suggestion
```

cips/cip-031.md
```diff
@@ -67,48 +66,83 @@ By integrating staking rewards directly into the lockup schedule, this proposal
...
+   The claimed rewards simply increase the total balance that will become available at the predefined unlock date.
 
+### Periodic Vesting Accounts
 
-```
+Periodic account creation will be removed, and update schedule will be a noop
+
+### Permanent Vesting Accounts
+
+Periodic account creation will be removed, and update schedule will be a noop
```
rootulp, line 105:
```suggestion
Permanent account creation will be removed, and update schedule will be a noop
```

cips/cip-031.md
```diff
@@ -67,48 +66,83 @@ By integrating staking rewards directly into the lockup schedule, this proposal
...
+
+    *NOTE: reward calculation is arbitrary, it is not based on Celestia's inflation.*
```
rootulp, line 107: this seems out of place
```suggestion
```

cips/cip-031.md
```diff
@@ -67,48 +66,83 @@ By integrating staking rewards directly into the lockup schedule, this proposal
...
+
+5. **Error Handling:**
+
+- Ensure sufficient balance checks before processing rewards.
+- Validate that the lockup period is still active.
+- Separate commission transfers from reward lockup updates.
```
rootulp, line 113: this isn't relevant because permanent vesting is a no-op. It should be removed from this section. It may be added to another section.

```suggestion
```

cips/cip-031.md
```diff
@@ -122,23 +156,59 @@ When upgrading to v4 we propose introducing a migration that will set the Parame
 
 # Backwards Compatibility
 
-All existing lockup accounts will be subject to the locking system. This change will take effect in v4, since this is a larger upgrade there is not a need for backwards compatible systems.
+Continuous Vesting, Delayed Vesting will implement the `UpdateSchedule` function allowing for the original amount to be added to. Periodic and Permanent will implement the design as a noop, and they will be disallowed from being created.
```
rootulp, line 159:
```suggestion
The continuous and delayed vesting account types will be updated to implement CIP-31. Specifically, they will implement the `UpdateSchedule` function so that the original vesting amount can account for staking rewards. The periodic and permanent account types will not support CIP-31. They will be disallowed from being created. 
```

[question] why do we need to disallow creating them? What about periodic and permanent vesting accounts that already exist?

cips/cip-031.md
```diff
@@ -122,23 +156,59 @@ When upgrading to v4 we propose introducing a migration that will set the Parame
...
-- **Redelegation with Rewards:**
-  - Ensure the hook triggers before redelegation and that rewards are claimed and added to lockup.
-- **Unbonding with Rewards:**
-  - Verify that the lockup schedule is updated correctly prior to unbonding.
 - **Validator Commission Path:**
   - Confirm that commission rewards are processed normally (non-locked) while staking rewards are added to the lockup schedule.
 
 # Reference Implementation
 
+The `VestingAccount` interface will be updated to account for an UpdateSchedule method which includes the amount to update the schedule with and teh blocktime at which it occured. We must pass in the blocktime in order to understand if the account is fully vested or not.
```
rootulp, line 172:
```suggestion
The `VestingAccount` interface will be updated to account for an UpdateSchedule method which includes the amount to update the schedule with and the blocktime at which it occured. We must pass in the blocktime in order to understand if the account is fully vested or not.
```

cips/cip-031.md
```diff
@@ -122,23 +156,59 @@ When upgrading to v4 we propose introducing a migration that will set the Parame
...
+	GetVestingCoins(blockTime time.Time) sdk.Coins
+
+	GetStartTime() int64
+	GetEndTime() int64
+
+	GetOriginalVesting() sdk.Coins
+	GetDelegatedFree() sdk.Coins
+	GetDelegatedVesting() sdk.Coins
+
+	// UpdateSchedule updates the original vesting amount for the account.
```
rootulp, line 207:
```suggestion
+	// UpdateSchedule updates the original vesting amount for the account.
```

cips/cip-031.md
tac0turtle: disallowing creation of them in order to simplify the protocol and due to proposing not implementing update schedule. Existing accounts, of which I couldnt find any, would still work, only the creation is disallowed

cips/cip-031.md
tac0turtle: updated the title but as the previous steps have numbers, breaking it here would cause confusion

cips/cip-031.md
rootulp: If existing accounts still work then I don't understand why celestia-app can't support creating new ones. 

Proposal: 
- Implement no-op `UpdateSchedule` for permanent and vesting accounts
- Preserve the ability to create new permanent and vesting accounts
- Clarify CIP-31 only applies to continuous and delayed vesting accounts

rootulp: Also given the most recent iteration of the implementation takes all staking rewards and updates the original vesting amount, it seems feasible to implement CIP-31 for periodic vesting accounts. AFAIU the complexity in the first implementation occurred when updating the amount to vest for various periods but that logic no longer needs to be added, right?

cips/cip-031.md
rootulp: If you want a title for the sections then please use a heading level with higher precedence than the H3s that follow. For example

```diff
- 4. **Mathematical Model & Example:**
+ ## Example Implementations

### Continuous Vesting Accounts

### Delayed Vesting Accounts

```

Currently this isn't formatted well. The rendered markdown looks like this

Screenshot 2025-04-07 at 11 51 02 PM

cips/cip-031.md
```diff
@@ -66,49 +65,76 @@ By integrating staking rewards directly into the lockup schedule, this proposal
...
+
+### Module Interaction
+
+The following sequence summarizes the module interactions:
+
+  ```mermaid
+  sequenceDiagram
+  participant DM as Distribution Module
+  participant AK as Account Keeper
+  participant LM as Lockup Module
```
rootulp, line 115: [nit] it's confusing to call this the lockup module when in code it is the vesting module

cips/cip-031.md
```diff
@@ -66,49 +65,76 @@ By integrating staking rewards directly into the lockup schedule, this proposal
...
+
+  rect rgb(0, 0, 139)
+    Note over DM: Account Type Check
+    DM->>DM: Is Lockup Account?
+  end
+
+  alt Is Lockup Account
+    DM->>LM: Get Lockup Type
+    DM->>LM: Calculate New Lockup Schedule
+    LM->>LM: Update Lockup Parameters
```
rootulp, line 129: What does this map to in code?

cips/cip-031.md
tac0turtle: for periodic this doesnt work as periods are proposed in json on creation. You need to update the periods as the periods contain the amount unlocked.

cmwaters (commented):
Thanks for the examples - they help explain how claiming works a lot.

I actually have a follow up question targeting the commission cap:
- Validators that currently have a commission of 100% won't have their commission affected right - just new validators.
- What about validators that change their commission? They won't be able to update their validator to a commission that is beyond the 25%. Is this reflected in the PR?

cips/cip-031.md
tac0turtle: i agree, when writing the CIP it was advised to use lockup nomenclature here even though its called vesting. I would prefer to keep it as is. on line 17 I added a note for this confusion.

tac0turtle:
> * Validators that currently have a commission of 100% won't have their commission affected right - just new validators.
correct
> * What about validators that change their commission? They won't be able to update their validator to a commission that is beyond the 25%. Is this reflected in the PR?

That is not reflected. I added in the latest commit a sentence on this

cips/cip-031.md
tac0turtle: amended to `#` instead to have a better flow as # and numbers throw things off

cips/cip-031.md
rootulp: Got it, thanks! Resolving.

rootulp (approved):

PR #312, opened 2025-06-16
cip to reduce unbonding period to ~14 days

cmwaters:


cips/cip-unbonding-period.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Backwards Compatibility
+
+This change introduces modifications that require careful coordination:
+
+1. **Network Upgrade Required**: The unbonding period change must be implemented through a coordinated network upgrade to ensure all validators adopt the new parameters simultaneously.
+
+2. **Existing Unbonding Delegations**: Delegations already in the unbonding process at the time of upgrade will continue under the previous 21-day period to avoid disrupting existing commitments.
+
+3. **Existing Governance Proposals to Modify Evidence Parameters**: Any existing proposals, though unlikely, must not be allowed to change the evidence params following the coordinated upgrade.
```
nashqueue, line 62: Are they modifiable, and are they changed to non-modifiable after the upgrade? Looking at the table of this CIP

PR #337, opened 2025-07-25
docs: cip-Privval Interface Extension for Arbitrary Message Signing

evan-forbes:
## Overview

adds CIP-40

### for editors
- [ ] assign CIP number and rename file
- [x] add to cips/README.md - to render on CIPs table
- [x] add to cips/SUMMARY.md - to render on the site

### DIFF
--- a/cips/cip-040.md
+++ b/cips/cip-040.md
@@ -123,7 +123,7 @@ This proposal is fully backwards compatible. Existing KMS implementations will c
 Applications using this interface MUST implement proper signature verification including:
 
 - Validating the signer's public key against the validator set
-- Verifying the chain_id matches the expected network
+- Verifying the `chain_id` matches the expected network
 - Confirming the sign bytes (`raw_message`) are correct, likely by generating the raw message itself.
 
 ## Reference Implementation


cips/cip-040.md
```diff
@@ -0,0 +1,142 @@
...
+    SignedVoteResponse       signed_vote_response        = 4;
+    SignProposalRequest      sign_proposal_request       = 5;
+    SignedProposalResponse   signed_proposal_response    = 6;
+    PingRequest              ping_request                = 7;
+    PingResponse             ping_response               = 8;
+    SignP2PMessageRequest    sign_p2p_message_request    = 9;
+    SignedP2PMessageResponse signed_p2p_message_response = 10;
+  }
+}
+```
```
rootulp, line 64: Consider using a diff block here to make it clear that the extension only includes two message types. The other fields already exist.

```suggestion
```diff
message Message {
  oneof sum {
    PubKeyRequest            pub_key_request             = 1;
    PubKeyResponse           pub_key_response            = 2;
    SignVoteRequest          sign_vote_request           = 3;
    SignedVoteResponse       signed_vote_response        = 4;
    SignProposalRequest      sign_proposal_request       = 5;
    SignedProposalResponse   signed_proposal_response    = 6;
    PingRequest              ping_request                = 7;
    PingResponse             ping_response               = 8;
+    SignP2PMessageRequest    sign_p2p_message_request    = 9;
+   SignedP2PMessageResponse signed_p2p_message_response = 10;
  }
}
```
```

File: cips/cip-040.md [hunk 2]
```diff
@@ -0,0 +1,142 @@
...
+| author | CHAMI Rachid ([@rach-id](https://github.com/rach-id)), Evan Forbes ([evan-forbes](https://github.com/evan-forbes)) |  
+| discussions-to | <https://forum.celestia.org/t/cip-40-privval-interface-extension-for-arbitrary-message-signing/2102> |
+| status | Draft |
+| type | Standards Track |
+| category | Interface |
+| created | 2025-07-25 |
+
+## Abstract
+
+This CIP proposes extending the CometBFT privval.Message interface to support signing arbitrary message hashes. This enhancement enables validators to use their consensus keys and existing Key Management Systems (KMS) to sign offchain protocol messages, facilitating the implementation of high-throughput and low-latency gossiping protocols while maintaining backwards compatibility.
```
rootulp, line 14: [no change needed][question] why message hashes instead of the raw message?

cips/cip-040.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Rationale
+
+### Design Decisions
+
+**Hash-based Signing**: The interface accepts pre-hashed messages rather than raw message content to:
+
+- Maintain consistent message size regardless of original content length
+- Ensure cryptographic properties of the signed data
+- Simplify KMS implementations by avoiding message content validation
```
rootulp, line 90: Does each validator get the raw message in addition to the hash of the message before signing? 

If not, is the validator blindly signing over a hash without access to the raw message?

cips/cip-040.md
rootulp: Disregard, this is elaborated on below.

cips/cip-040.md
```diff
@@ -0,0 +1,142 @@
...
+
+KMS implementations SHOULD implement rate limiting for P2P signing requests to prevent resource exhaustion attacks against the signing infrastructure.
+
+### Key Compromise
+
+In the event of key compromise, the impact extends beyond consensus to any protocols using this signing interface. Monitoring systems SHOULD track unusual signing patterns across both consensus and P2P message types.
+
+## Reference Implementation
+
+A reference implementation is available in CometBFT PR [#5138](https://github.com/cometbft/cometbft/pull/5138). This implementation has been tested on multiple private testnets and the Mamo testnet, demonstrating practical viability.
```
rootulp, line 136: This PR appears to allow signing over arbitrary bytes not just hashes so it seems slightly divergent from this spec.

Also the requests are named differently:
- sign_raw_bytes_request in the CometBFT PR
-  sign_p2p_message_request in this spec

rootulp (approved):

cips/cip-040.md
evan-forbes: great idea! fixed in https://github.com/celestiaorg/CIPs/pull/337/commits/3b766642f9653d1fc34cd8d9abf203dab20a6254

cips/cip-040.md
```diff
@@ -0,0 +1,118 @@
...
+| author | CHAMI Rachid ([@rach-id](https://github.com/rach-id)), Evan Forbes ([evan-forbes](https://github.com/evan-forbes)) |
+| discussions-to | <https://forum.celestia.org/t/cip-40-privval-interface-extension-for-arbitrary-message-signing/2102> |
+| status | Draft |
+| type | Standards Track |
+| category | Interface |
+| created | 2025-07-25 |
+
+## Abstract
+
+This CIP proposes extending the CometBFT privval.Message interface to support signing arbitrary messages. This enhancement enables validators to use their consensus keys and existing Key Management Systems (KMS) to sign offchain protocol messages, facilitating the implementation of high-throughput and low-latency gossiping protocols while maintaining backwards compatibility.
```
rach-id, line 14: proposition to have code in `these` blocks. If you agree, maybe we should change all occurrences and also the chainID, uniqueID, etc
```suggestion
This CIP proposes extending the CometBFT `privval.Message` interface to support signing arbitrary messages. This enhancement enables validators to use their consensus keys and existing Key Management Systems (KMS) to sign offchain protocol messages, facilitating the implementation of high-throughput and low-latency gossiping protocols while maintaining backwards compatibility.
```

cips/cip-040.md
rach-id: Yes, we should use the raw messages because ed25519 is vulnerable for hashes

cips/cip-040.md
```diff
@@ -0,0 +1,118 @@
...
++    SignRawMessageRequest    sign_raw_message_request    = 9;
++    SignedRawMessageResponse signed_raw_message_response = 10;
+  }
+}
+```
+
+### Field Specifications
+
+- `chain_id`: The chain identifier to prevent cross-chain signature reuse.
+- `raw_bytes`: The digest that is signed over.
```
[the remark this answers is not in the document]
evan-forbes: I see good point, we can reword this field doc to clarify that this digest is not the entire digest and that the unique id is also included.

we can also reword the unique field id to mention that its not really optional, since its included in the sign bytes

> **Unique ID Field**: The optional unique_id field allows protocols to differentiate between different message types or contexts without requiring separate interface extensions.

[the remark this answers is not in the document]
evan-forbes: see

```
message SignRawBytesRequest {
  string chain_id  = 1;
  bytes  raw_bytes = 2;
  string unique_id = 3;
}
```

I forgot that uniqueID is actually optional since its a string and therefore can be "", same with chain-id

> a rogue consensus node could make the KMS double sign votes and proposals.

then this makes sense to me if we're using strings across the board cc @rach-id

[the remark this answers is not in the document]
evan-forbes: the fix here is to not make unique-id optional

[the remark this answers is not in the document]
rach-id: That's a very good point 👍 making it optional will have the same vulnerability. Maybe we should generate a nonce on the KMS, prepend it before signature, return it along with the signed bytes. This would guarantee uniqueness and add random data even if the payload is trying to mimic another message.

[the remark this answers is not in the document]
evan-forbes: see https://github.com/celestiaorg/CIPs/pull/337/commits/89e64eb68916864bb320f23d323b6e54d091a9ef

cips/cip-040.md
```diff
@@ -0,0 +1,136 @@
...
+
+The proposed extension addresses this gap by allowing validators to sign arbitrary messages while maintaining the security properties of their existing key management infrastructure.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Protocol Buffer Definitions
+
+The following message types SHALL be added to the privval.Message interface:
```
jcstein, line 30:
```suggestion
The following message types SHALL be added to the `privval.Message` interface:
```

cips/cip-040.md
```diff
@@ -0,0 +1,136 @@
...
+message SignedRawBytesResponse {
+  bytes             signature = 1;
+  RemoteSignerError error     = 2;
+}
+
+```
+
+### Interface Extension
+
+The privval.Message interface SHALL be extended to include the new message types:
```
jcstein, line 48:
```suggestion
The `privval.Message` interface SHALL be extended to include the new message types:
```

cips/cip-040.md
```diff
@@ -0,0 +1,136 @@
...
+4. Double-signing protection is NOT REQUIRED for raw message signing operations.
+5. Convert each component to its byte representation as specified above
+6. Concatenate the byte sequences directly without delimiters
+7. Sign the resulting byte array
+
+## Rationale
+
+### Design Decisions
+
+**Chain ID Inclusion**: Including the chain_id prevents signature reuse across different networks.
```
jcstein, line 104:
```suggestion
**Chain ID Inclusion**: Including the `chain_id` prevents signature reuse across different networks.
```

cips/cip-040.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Security Considerations
+
+### Signature Verification
+
+Applications using this interface MUST implement proper signature verification including:
+
+- Validating the signer's public key against the validator set
+- Verifying the chain_id matches the expected network
+- Confirming the sign bytes (raw_message) are correct, likely by generating the raw message itself.
```
jcstein, line 126:
```suggestion
- Confirming the sign bytes (`raw_message`) are correct, likely by generating the raw message itself.
```

jcstein (approved):
LGTM aside from unresolved comments from @rach-id and @evan-forbes

cips/cip-040.md
```diff
@@ -0,0 +1,137 @@
...
+This proposal is fully backwards compatible. Existing KMS implementations will continue to function normally, as the new message types use previously unused field numbers in the protobuf oneof union. That being said, all repos that import the interface MUST update their implementions to at least use a noop.
+
+## Security Considerations
+
+### Signature Verification
+
+Applications using this interface MUST implement proper signature verification including:
+
+- Validating the signer's public key against the validator set
+- Verifying the chain_id matches the expected network
```
jcstein, line 126:
```suggestion
- Verifying the `chain_id` matches the expected network
```

cips/cip-040.md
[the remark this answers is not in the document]
evan-forbes: the protobuf thing is interesting, and might also apply to nonces then as well if they are an issue no?

either way, the chain-id is being added by the kms, so as long as that's the case I think we're fine and the original concern also isn't valid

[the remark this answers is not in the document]
rach-id: hum, I think we're fine without changing the unique_id with an int. Check the sign bytes:

image

the unique_id is required, and also we add a domain prefix to the sign bytes. So the issue shouldn't happen at all. Apologies I thought I remembered the implementation, but I forgot the prefix thing

cips/cip-040.md
```diff
@@ -0,0 +1,137 @@
...
+- `signature`: The resulting signature bytes from the signing operation.
+- `error`: Error information if the signing operation fails.
+
+### Sign Bytes Construction
+
+The actual bytes that are signed MUST be constructed by concatenating:
+
+```text
+sign_bytes = chain_id + unique_id + raw_bytes
+```
```
cmwaters, line 81: 👍 thanks

cips/cip-040.md
```diff
@@ -0,0 +1,118 @@
...
+
+**No Double-Signing Protection**: Unlike consensus messages, raw messages do not require double-signing protection as they do not affect chain safety.
+
+### Alternative Approaches Considered
+
+**Separate Interface**: Creating a completely separate interface was considered but rejected to maintain consistency with existing KMS integrations and avoid fragmenting the signing infrastructure.
+
+## Backwards Compatibility
+
+This proposal is fully backwards compatible. Existing KMS implementations will continue to function normally, as the new message types use previously unused field numbers in the protobuf oneof union. That being said, all repos that import the interface MUST update their implementions to at least use a noop.
```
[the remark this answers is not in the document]
cmwaters: I see, perhaps they can continue to run and they just have noisy logs

cmwaters (approved):
Thanks for addressing my suggestions. LGTM!

PR #340, opened 2025-07-28
CIP to reduce issuance

cmwaters:
Proposes changes to the protocol that would halve issuance from 5% in v4 to 2.5% in v5 and double minimum commission from 5% to 10%

### For editors
- [ ] assign CIP number, rename file
- [ ] add to README.md - to include on homepage
- [ ] add to SUMMARY.md - to render on site


cips/cip-reduce-issuance.md
```diff
@@ -0,0 +1,125 @@
...
+| **v4** | 5.00% | 5.00% | After CIP-29 reduction |
+| **v5** | 4.67% | 2.50% | **Proposed CIP** |
+| **2** | 4.35% | 2.33% | Regular annual disinflation applied (6.7%) |
+| **3** | 4.06% | 2.17% | Regular annual disinflation applied (6.7%) |
+| **4** | 3.79% | 2.02% | Regular annual disinflation applied (6.7%) |
+| **5** | 3.54% | 1.88% | Regular annual disinflation applied (6.7%) |
+| **6** | 3.30% | 1.75% | Regular annual disinflation applied (6.7%) |
+| **7** | 3.08% | 1.63% | Regular annual disinflation applied (6.7%) |
+| **8** | 2.87% | 1.52% | Regular annual disinflation applied (6.7%) |
+| **9** | 2.68% | 1.50% | Target inflation reached |
```
rootulp, line 89: [optional] CIP-29 has this table: https://github.com/celestiaorg/CIPs/blob/ab63f71b4a2ef8477ea1be1695e1df585768b11e/cips/cip-029.md?plain=1#L45-L66 

1. Can the table in this PR expand to 18 or 20 years?
2. [optional] is it worth expanding the decimal places so that the current inflation in this table matches the proposed inflation from the previous table?

cips/cip-reduce-issuance.md
```diff
@@ -0,0 +1,125 @@
...
+
+**Note**: The table assumes this CIP would be implemented less than 2 years after genesis.
+
+## Backwards Compatibility
+
+This change introduces modifications that require careful coordination:
+
+1. **Network Upgrade Required**: The inflation and commission changes must be implemented through a coordinated network upgrade to ensure all validators adopt the new parameters simultaneously.
+
+2. **Validator Commission Adjustments**: This will need to be done as a governance proposal and can happen prior of after the inflation change.
```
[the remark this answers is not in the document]
rootulp: [nit][optional] we can keep the parameter governance modifiable and change it in an upgrade handler form v4 -> v5

rootulp (approved):

cips/cip-reduce-issuance.md
```diff
@@ -0,0 +1,125 @@
+| cip | XX (assigned by Editors) |
+| - | - |
+| title | Further reduce issuance to 2.5% and increase minimum commission to 10% |
+| description | Reduce inflation to 2.5% and increase minimum validator commission to 10% to improve TIA's suitability for financial applications |
+| author | Callum Waters (@cmwaters), Nashqueue (@nashqueue) |
+| discussions-to | URL |
+| status | Draft |
+| type | Standards Track |
+| category | Core |
+| created | 2025-01-02 |
```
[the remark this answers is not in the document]
jcstein: I think we could add that, or does it depend on it? https://cips.celestia.org/cip-001.html?highlight=depends-on#requires-header

cips/cip-reduce-issuance.md
jcstein: I think it should extend, too. 20 seems safe.

cips/cip-reduce-issuance.md
```diff
@@ -0,0 +1,125 @@
+| cip | XX (assigned by Editors) |
+| - | - |
+| title | Further reduce issuance to 2.5% and increase minimum commission to 10% |
+| description | Reduce inflation to 2.5% and increase minimum validator commission to 10% to improve TIA's suitability for financial applications |
+| author | Callum Waters (@cmwaters), Nashqueue (@nashqueue) |
+| discussions-to | <https://forum.celestia.org/t/cip-reduce-issuance-to-2-5-and-increase-minimum-commission-to-10/2103> |
+| status | Draft |
+| type | Standards Track |
+| category | Core |
+| created | 2025-07-28 |
```
jcstein, line 10:
```suggestion
| created | 2025-07-28 |
| requires | [CIP-29](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-029.md) |
```

cips/cip-reduce-issuance.md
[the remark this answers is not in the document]
jcstein: for now https://github.com/celestiaorg/CIPs/pull/340/commits/4cc06b795893a4aea2638dd4ad7da09ac9e9a195 and please unresolve if you disagree!

cips/cip-reduce-issuance.md
jcstein: I think extending decimals also makes sense

cmwaters: But nothing else happens after year 9. There's no new information to convey when target inflation is reached

NRDLabs:
Currently, a validator earns approximately $4,108 per month with a 10% commission, 3 million TIA delegated, a token price of $1.6, and an APR of 10.27%. We are aware that in the coming month, 43.5 million TIA purchased from Polychain will return to staking, which will increase the staking ratio to ~51.4% and reduce the APR to ~9.72%, bringing validator earnings down to approximately $3,888 per month.

Validators must cover the costs of bare-metal servers, monitoring systems, IBC relayers, backup nodes, alerting systems, key management, and security infrastructure. To maintain decentralization, many operators choose to run infrastructure in geographically diverse and less common data centers, which significantly increases expenses. At current levels, profitability is already near the margin.

If the inflation is reduced to 2.5%, with the same staking ratio, the APR would drop to ~4.86%, resulting in validator income of only ~$1,944 per month — half the current level. In such conditions, many validators will be forced to abandon high-quality infrastructure in favor of cheaper cloud setups, which could compromise the reliability and decentralization of the network.

Furthermore, in the recent forum discussion Proof of Governance as the endgame for LSTs, Adlerjohn emphasizes the importance of maintaining validator incentives, as a key factor in ensuring network security. Sustaining adequate validator income is not just an economic issue — it’s fundamental to the security and robustness of consensus.

Reducing inflation month after month is not a sustainable strategy. It might make sense under different market conditions, but certainly not right now, when the network is still growing and infrastructure costs remain high. We believe this proposal does not represent an improvement and that changes to inflation must be considered more cautiously, with long-term validator sustainability in mind.

PR #366, opened 2025-11-06
docs(cip-19): add missing shwap IDs along with the containers

vgonkivs:
Added missing shwap types:
* IDs - NamespaceDataID, RangeNamespaceDataID
* Containers - NamespaceData, RangeNamespaceData

### DIFF
--- a/cips/cip-019.md
+++ b/cips/cip-019.md
@@ -364,8 +364,8 @@ RangeNamespaceDataID identifiers formatted as shown below:
 ```text
 RangeNamespaceDataID {
     EdsID;
-    From: u16;
-    To: u16;
+    From: u32;
+    To: u32;
 }
 ```
 


cips/cip-019.md
```diff
@@ -313,6 +314,99 @@ Namespace data may span over multiple rows, in which case all the data is encaps
...
+
+**From**: A uint16 representing the first index of the sampled share within the ODS. The 16 bit
+limit fits data squares up to 2TB. It MUST stay within the number of original shares(1st quadrant) of the EDS.
+
+**To**: A uint16 representing the last index of the sampled share within the ODS(exclusively). The 16 bit
+limit fits data squares up to 2TB. It MUST stay within the number of original shares(1st quadrant) increased by one of the EDS.
+
+#### RangeNamespaceData Container
+
+RangeNamespaceData containers encapsulate user-submitted data under speicific coordinates [from;to] within a [DataSquare][square].
```
zvolin: we could use math notation, to express the exclusiveness of `to`
```suggestion
RangeNamespaceData containers encapsulate user-submitted data under speicific coordinates [from,to) within a [DataSquare][square].
```

cips/cip-019.md
```diff
@@ -313,6 +314,99 @@ Namespace data may span over multiple rows, in which case all the data is encaps
...
+**Shares**: A variable size [RowShares](#rowshares) array representing data of a range in a single [DataSquare][square].
+
+**FirstIncompleteRowProof**: An [NMT][nmt] proof for the first row, present when:
+
+- The range starts at a non-zero share index, OR
+- In single-row ranges, when the range is not full (less than the number of shares in the row)
+
+**LastIncompleteRowProof**: An [NMT][nmt] proof for the last row in multi-row ranges, present when the range ends before the row's final share.
+
+Both Proofs can be empty.
```
zvolin: how can someone prove the data? `RowShares` don't carry any proofs.
When having only the ODS part of the row, the only thing that comes to my mind is that one needs to recompute the parity data, create an Nmt of the whole row and then compare the root with the DAH? Sounds like quite a lot of computation, especially for bigger squares

cips/cip-019.md
```diff
@@ -313,6 +314,99 @@ Namespace data may span over multiple rows, in which case all the data is encaps
...
+
+The fields with validity rules that form RangeNamespaceDataID are:
+
+[**EdsID**](#edsid): A EdsID of the RangeNamespaceData container. It MUST follow [EdsID](#edsid) formatting and field validity rules.
+
+**From**: A uint16 representing the first index of the sampled share within the ODS. The 16 bit
+limit fits data squares up to 2TB. It MUST stay within the number of original shares(1st quadrant) of the EDS.
+
+**To**: A uint16 representing the last index of the sampled share within the ODS(exclusively). The 16 bit
+limit fits data squares up to 2TB. It MUST stay within the number of original shares(1st quadrant) increased by one of the EDS.
```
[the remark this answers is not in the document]
walldiss: This is really good catch! Yeah, your calculations is correct and using u16 for coordinates has limit of requesting ranges only from <32MB block. 

@vgonkivs Lets bump it to u32 in spec and make an issue use usage of 32 bits in shrex version of Range request.

jcstein:
once resolved, we can merge 🫡

PR #372, opened 2026-01-15
docs(cip-43): protocol-injected tx for dashboard-compatible fee forwarding

Manav-Aggarwal:
## Summary
Update CIP-43 to describe the protocol-injected transaction design for dashboard-compatible fee forwarding.

## Changes
- Add `MsgForwardFees` message type specification
- Document PrepareProposal injection mechanism
- Document ProcessProposal strict validation
- Add dashboard compatibility rationale
- Update security considerations for unsigned transactions

## Reference Implementation
celestiaorg/celestia-app#6412

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cips/cip-043.md
+++ b/cips/cip-043.md
@@ -1,7 +1,7 @@
 | cip | 43 |
 | - | - |
 | title | Fee address module |
-| description | Contribute TIA tokens to protocol fee revenue via a vanity fee address |
+| description | Forward TIA tokens to protocol revenue via a dedicated fee address |
 | author | Manav Aggarwal ([@Manav-Aggarwal](https://github.com/Manav-Aggarwal)) |
 | discussions-to | <https://forum.celestia.org/t/cip-add-burn-module/2216> |
 | status | Draft |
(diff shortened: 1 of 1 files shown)


cips/cip-043.md
```diff
@@ -11,96 +11,138 @@
...
-The `FeeAddressIBCMiddleware` intercepts inbound IBC transfers (`OnRecvPacket`). If the receiver is the fee address:
+When preparing a block proposal, the block proposer:
 
-- **Allowed**: Native `utia` (including utia returning from other chains with IBC prefixes like `transfer/channel-0/utia`)
-- **Rejected**: Foreign tokens (e.g., `uosmo`, `uatom`, or any non-utia denom)
+1. Queries the fee address `utia` balance
+2. If balance is non-zero:
+   - Creates a `MsgForwardFees` transaction
+   - Sets the transaction fee to the **entire fee address balance**
+   - Sets gas limit to 50,000 (minimal, no computation)
```
rootulp: What if the entire fee address balance is so small (e.g. 1 utia) that this transaction doesn't exceed the network min gas price

cips/cip-043.md
```diff
@@ -109,41 +151,59 @@ This module introduces no new parameters.
 
 ## Rationale
 
-The fee address approach offers several advantages over a dedicated message:
+### Dashboard Compatibility
+
+Blockchain analytics dashboards (e.g., Blockworks, Celenium) calculate protocol revenue by aggregating transaction fees from the `tx.AuthInfo.Fee` field. A naive design using direct bank transfers in EndBlocker would not appear in these revenue calculations.
```
rootulp: Is it a requirement that the protocol be designed to accommodate these dashboards?

Naively it seems far easier to modify the dashboard logic than the protocol logic. This design is significantly more complex than the two previous designs so if it's possible for the dashboards to be updated, I am in favor of the earlier less complex protocol designs.

Manav-Aggarwal: > I assume we're already coordinating with dashboards by asking them how they calculate existing protocol revenue. How do they currently calculate protocol revenue and fees? Is it consistent across all dashboards?

We're not coordinating with all dashboards, so a reasonable assumption to make here is that dashboards are summing up transaction fees across all transactions to calculate protocol revenue and fees.

> The original CIP (with MsgBurn) also enabled dashboards to query for protocol revenue and fees. Given all the info I've seen so far, it seems more straight forward to introduce a MsgBurn message and for dashboards to query total supply via the existing bank total supply method.

This would require coordinating with all dashboards to add functionality to them which is not under our control and we've seen unwillingness from dashboards like Defillama in the past to coordinate. Purely introducing a MsgBurn message type also requires a custody address like a multisig to be setup on Celestia to make that message type which is undesirable hence why the introduction of an address to which sends are considered fees for the protocol.

> I don't quite follow the rationale for making this CIP more and more complicated with each iteration.

At each iteration, I've gotten feedback on issues with the previous design which has led to UX improvements to accommodate simple sends and maintaining compatibility with dashboards. Happy to also discuss the progression over a call

PR #396, opened 2026-04-27
[codex] Finalize v7 CIPs and add v8 CIP

jcstein:
## Summary

- Keep CIP-47 as the Hibiscus (v7) Network Upgrade and note that v7 did not activate on Mainnet Beta after an `x/forwarding` bug was discovered during testnet rollout.
- Add CIP-49 as the separate Draft v8 Network Upgrade Meta CIP, carrying forward CIP-44, CIP-45, and CIP-46 without reusing the Hibiscus name.
- Move CIP-44, CIP-45, CIP-46, and CIP-47 from Last Call to Final and remove their last-call-deadline metadata.
- Add CIP-49 to the CIP overview table and mdBook summary.

## Validation

- `npx markdownlint-cli@0.39.0 --config .markdownlint.yaml cips/cip-044.md cips/cip-045.md cips/cip-046.md cips/cip-047.md cips/cip-049.md cips/README.md cips/SUMMARY.md`
- `npx markdownlint-cli@0.39.0 --config .markdownlint.yaml cips/cip-049.md cips/README.md cips/SUMMARY.md`
- `git diff --check`

### DIFF
--- a/cips/README.md
+++ b/cips/README.md
@@ -97,7 +97,7 @@ Read [CIP-1](./cip-001.md) for information on the CIP process.
 | [44](./cip-044.md) | Adjust Validator Commission Bounds                                           | [@cmwaters](https://github.com/cmwaters)                                                                                                                                     |
 | [45](./cip-045.md) | Forwarding Module                                                            | [@Manav-Aggarwal](https://github.com/Manav-Aggarwal)                                                                                                                         |
 | [46](./cip-046.md) | ZK Interchain Security Module                                                | [@Manav-Aggarwal](https://github.com/Manav-Aggarwal)                                                                                                                         |
-| [47](./cip-047.md) | Hibiscus (v7) Network Upgrade                                                | [@rootulp](https://github.com/rootulp)                                                                                                                                       |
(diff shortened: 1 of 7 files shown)


rootulp (approved):
It may be confusing to call v8 Hibiscus because v7 was Hibiscus. I'm not opinionated on the upgrade names and internally refer to them as v7 and v8 to avoid confusion.

jcstein:
should we make a new CIP for v8? and call it v8, and Hibiscus (v7) just "skips" mainnet?

rootulp:
That works too and seems slightly preferable.

jcstein:
agreed. i think maybe it would be best to edit v7 hibiscus to only include what was actually in v7. i'll need to think on this a bit more and make a FLUP with the "real" v8

rootulp:
v7 included all those CIPs, we just prevented it from shipping to Mainnet because the x/forwarding bug. v8 also includes all those CIPs.

rootulp (approved):
Thanks, I like this more

jcstein:
you're welcome, same here

